PR #185, opened 2014-11-20
Allowing pip-install to have access to config vars

pngdwnld:
- Currently, `pip-install` does not have access to config vars. 
- This PR is particularly useful when installing a pip package that requires an apt-package. For example, `scipy` requires `libatlas-dev` -- in this case, I can use `heroku-buildpack-multi` and `heroku-buildpack-apt`, and config vars `BLAS=/app/.apt/usr/lib/libblas/libblas.a` would let the pip package `scipy` knows the location of the static library during `pip-install`.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -166,7 +166,7 @@ source $BIN_DIR/steps/pylibmc
 source $BIN_DIR/steps/cryptography
 
 # Install dependencies with Pip.
-source $BIN_DIR/steps/pip-install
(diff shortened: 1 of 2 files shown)


dzuelke:
This change runs the `pip install` in a subshell, where a failure is not detected by `set -e`: http://fvue.nl/wiki/Bash:_Error_handling#Solutions

As a result, if the `pip install` fails, the build still succeeds.

PR #193, opened 2015-01-08
enable custom requirements-uninstall.txt

miraculixx:
This change enables to run a custom pip uninstall of immediately before pip-install, e.g. to refresh some packages:

*Usage*
* add `requirements-uninstall.txt`
* git push

*How it works*
`steps/pip-install` will detect `requirements-uninstall.txt`, process all requirements in it (including support for -r inclusion of other requirements). Immediately afterwards, continues with `pip install`.

### DIFF
--- a/bin/steps/pip-install
+++ b/bin/steps/pip-install
@@ -8,6 +8,11 @@ puts-step "Installing dependencies with pip"
 [ ! "$FRESH_PYTHON" ] && bpwatch start pip_install
 [ "$FRESH_PYTHON" ] && bpwatch start pip_install_first
 
+# if user wants to uninstall some packages first do so
+if [[ -f requirements-uninstall.txt ]]; then
+  pip uninstall -y -r requirements-uninstall.txt
+fi
+
 /app/.heroku/python/bin/pip install -r requirements.txt --exists-action=w --src=./.heroku/src --allow-all-external  | cleanup | indent
 
 # Smart Requirements handling


kennethreitz:
The python buildpack already automatically detects packages that have been removed from requirements.txt, and will uninstall them automatically.

miraculixx:
Yes, yet that's not the point of this PR. This PR is about reinstalling packages that continue to stay in requirements.txt and thus are not detected by the stale mechanism.

The specific scenario this PR solves is this, which is not at all addressed by the current buildpack.

requirements.txt contains the following
-r requirements-public.txt
-r requirements-custom.txt

In -public, packages from the public repository are installed. In -custom, packages from a private repository are are installed. There are circumstances when you want to uninstall (and possibly reinstall the packages in -custom), e.g. to get a new version during a development/testing deploy. Of course we all know that you're supposed to bump your version when you start development of the new version, but we also all know that this is not realistic in practice.  This PR is making this scenario practicable.   

Besides it is a common problem:
* http://stackoverflow.com/questions/15046378/manually-remove-python-package-on-heroku
* http://stackoverflow.com/questions/8937905/how-to-pip-uninstall-with-virtualenv-on-heroku-cedar-stack

kennethreitz:
Can you be specific about a use-case that you are solving for? I understand your thought process, but I don't understand how this necessary.

PR #195, opened 2015-01-23
Add --upgrade to pip install

migurski:
I’m building an app with dependencies defined in `setup.py` using [install_requires](https://pythonhosted.org/setuptools/setuptools.html#declaring-dependencies) to list them. In my `requirements.txt`, I have a line with `.` on it that installs my main module and all its dependencies.

Without the `--upgrade` option, pip does not recognize the addition of new requirements in builds over time and incorrectly skips them. This patch fixes the bug and correctly installs locates all installation requirements that may have changed in a new push to Heroku.

### DIFF
--- a/bin/steps/pip-install
+++ b/bin/steps/pip-install
@@ -8,7 +8,7 @@ puts-step "Installing dependencies with pip"
 [ ! "$FRESH_PYTHON" ] && bpwatch start pip_install
 [ "$FRESH_PYTHON" ] && bpwatch start pip_install_first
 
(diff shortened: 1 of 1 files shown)


kennethreitz:
You should be doing `-e .` instead.

The Python buildpack actually automatically does this if you don't have a `requirements.txt` file present but do have a `setup.py`. Nice little trick :)

migurski:
Thank you! How can I do `-e .` with the existing buildpack + requirements.txt?

kennethreitz:
Just add that anywhere in your `requirements.txt` file.

migurski:
[Ah so](https://pip.pypa.io/en/latest/reference/pip_install.html#requirements-file-format), thanks.

PR #196, opened 2015-02-06
extend LIBRARY_PATH

t-io:
extend `LIBRARY_PATH` to keep paths from previous installed buildpacks. 
E.g __heroku-buildpack-apt__ does finally something like this:
`export | grep -E -e ' (PATH|LD_LIBRARY_PATH|LIBRARY_PATH|INCLUDE_PATH|CPATH|CPPPATH|PKG_CONFIG_PATH)='  > "$LP_DIR/export"`


kennethreitz:
This seems harmless enough. Hope I'm right.

Issue #202, opened 2015-03-12
pypy-2.5.0


brizjin:
Please, add support of pypy 2.5.0

cclauss:
This one can be closed.  https://github.com/heroku/heroku-buildpack-python/blob/master/builds/runtimes/pypy-2.5.1

Issue #197, opened 2015-02-10
pypy support


tardyp:
It looks like there is an access denied on pypy runtime:
http://lang-python.s3.amazonaws.com/cedar/runtimes/pypy-2.4.0

Moreover, pypy 2.5.0 has been out for some time now.

Please enable faster python on heroku. :)

cclauss:
This one can be closed.  https://github.com/heroku/heroku-buildpack-python/blob/master/builds/runtimes/pypy-2.5.1

PR #207, opened 2015-03-26
pypy-2.5.1

cclauss:
Issue: Execute bit is NOT yet set!

Should close issue #202 and part of issue #197


kennethreitz:
@cclauss why are you not setting execute bit? I'd love to merge these but never will if that's the case :)

cclauss:
Hi Kenneth,

    It has to do with the way that I create these pull requests... I paste them into the GitHub webpage interface.  I have not seen a way to set file permission bits via the GitHub web interface. When I get back to my desktop, I will try to resubmit them normal way. 

Thx, CCC

PR #211, opened 2015-04-02
Prefer a 'manage.py' that exists higher up in the directory hierarchy

ollanta:
If you had multiple files named `manage.py` (for example if you have other django projects as submodules), the old collectstatic could be trying to run one of those files instead of the main one.

This change sorts `manage.py`s by their depth in the directory hierarchy, so that the shallowest one is chosen.


kennethreitz:
this looks like a great change, thanks!

PR #216, opened 2015-04-23
Add symlinks to common version abbreviations

DavidJFelix:
I'm not entirely sure if this will work or if these values need to be transfered to a heroku backend elsewhere also.
My reasoning behind this change is that some people (at least me) don't care about minor versions and don't write code that will break on minor versions.
By symlinking the versions, you can allow people to say "I'm sure it will only work in python-2" and always get the newest python 2 or "I'm sure it only works in python 3.4". 

Just food (and a PR) for thought.


kennethreitz:
I like this idea, but there are a few things that prevent me from merging this:

- a lot of hard-to-track builds would have to be made every time there's a new release. 
- the way app version-tracking currently works, you wouldn't get upgrades automatically.

A better approach would be to add these links on the server side (s3), where the packages themselves live. 

So, we'd have to give these links a way to "resolve" (like a redirect) on the server side, and this would be done with S3.

DavidJFelix:
Yeah, I wasn't sure if adding them here would really do anything.

DavidJFelix:
@kennethreitz so what do you think we do with this PR. Close it?

kennethreitz:
for now, yes

PR #217, opened 2015-04-24
Detect pyOpenSSL

stevo550:
Some people use pyOpenSSL (it is the literal name, e.g.: https://pypi.python.org/pypi/pyOpenSSL) so how about we just detect both?


kennethreitz:
:sparkles: :cake: :sparkles:

PR #218, opened 2015-04-27
Improve detect script: /bin/sh, look for Python/Django

blueyed:
Without a requirements.txt and setup.py file in the root, it now will look for `manage.py` to detect a Django app.

This would allow for moving your requirements.txt file in the bin/pre_compile step.

### DIFF
--- a/bin/detect
+++ b/bin/detect
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 
 # This script serves as the
 # [**Python Buildpack**](https://github.com/heroku/heroku-buildpack-python)
--- a/bin/detect
+++ b/bin/detect
@@ -8,9 +8,8 @@
 # adapter between a Python application and Heroku's runtime.
 
 # ## Usage
-# Compiling an app into a slug is simple:
 #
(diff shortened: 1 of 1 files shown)


kennethreitz:
This is undesired behavior — all apps are simply Python apps. We used to detect Python/Django, but this is no longer useful or necessary.

blueyed:
Ok.
This could mean to only change the output in that case then.

The main point here is to also look for manage.py, and there are other improvements in there that could be cherry-picked.

kennethreitz:
All Django apps should also have a `requirements.txt` file, so this isn't a needed change.

blueyed:
Yes, they should have a requirements.txt, but it's often not in the root and/or would not be for Heroku, e.g. if you're using a "requirements" folder, with "requirements/heroku.txt" for Heroku.
There have been several issues and requests to make this more flexible already.

But even when this would still not be accepted, there are other commits in this PR.
What do you think about them?
Should I create separate PRs for them?

kennethreitz:
Every app should have exactly one `requirements.txt` file, not multiple.

blueyed:
In general you have different requirements for dev/local, production and testing.

PR #222, opened 2015-05-12
Optionally split src_dir and build_dir

Stibbons:
Hello

Here is the modification I have done to python buildpack in order to build my project. I use a modified version of heroku-buildpack-multi that support building projects that are in sub folders to the build directory (https://github.com/Stibbons/heroku-buildpack-multi-subdir).

In other word, I simply allow to build outside of the source directory. build_dir is where the .heroku directory will be generated. src_dir is where the requirements.txt file is. Of course, this is optional, the default is still build_dir = src_dir.

I add also to fix an issue regarding libffi which wasn't working. To do so, I've imported a modification from https://github.com/mfenniak/heroku-buildpack-python-libffi/blob/master/bin/steps/libffi.
I think this worth being generalized.

Tell me what you thing about this


kennethreitz:
Thank you very much! This is very cool.

Unfortunately, this is not something that will be merged into the official buildpack at this time.

PR #226, opened 2015-05-30
Set xtrace if $BUILDPACK_XTRACE set

msabramo:
For debugging or for folks who are curious about how the buildpack works.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -7,6 +7,8 @@
 # Fail fast and fail hard.
 set -eo pipefail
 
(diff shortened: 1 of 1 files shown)


kennethreitz:
can you tell me about xtrace?

msabramo:
Yeah, `set -o xtrace` is the long form of `set -x` which perhaps you might recognize better, but basically it just makes bash print out every command that it's executing with each line prefixed with "+ ". 

I was going to use this to better understand how this buildpack works and thought it could be generally useful.

kennethreitz:
This seems nice. Thanks.

PR #227, opened 2015-05-31
Break pip install command on to multiple lines

msabramo:
For better readability


kennethreitz:
i prefer the current method, but thanks!

PR #238, opened 2015-07-24
Update Readme.md

mattgraham:
Updated readme with official buildpack image.

python


kennethreitz:
@mattgraham would it be possible to get an updated version of this image?

> The official Heroku buildpack for python apps.

should be

> The official Heroku buildpack for Python apps.

PR #245, opened 2015-09-14
Run pip with the --upgrade flag

joshfriend:
This way, the app is always built the same way as if it was from scratch.

I think it's undesirable to have the app build differently if it is built from scratch or re-built from a previous push. Pushing the same code to a new app and an existing app should have the same result.

Personally, I've run into issues where a dependency becomes a bit outdated, it becomes difficult to force Heroku to update it. You have to:

1. Install the [heroku-repo](https://github.com/heroku/heroku-repo) extension
2. Run the `purge_cache` task
3. Run the `reset` task
4. Re-push the code to the app


kennethreitz:
we encourage all users to always be using pinned depenencies

PR #251, opened 2015-09-29
Permit $PATH from prior buildpacks.

ojacobson:
Multiple buildpack support means that prior buildpacks can export extra variables for use by later buildpacks. In particular, the `nodejs` buildpack, which is used for many asset pipelines, exports `node` and `npm` on `$PATH`.

### DIFF
--- a/bin/utils
+++ b/bin/utils
@@ -76,7 +76,7 @@ deep-rm() {
 sub-env() {
 
   WHITELIST=${2:-''}
-  BLACKLIST=${3:-'^(GIT_DIR|PYTHONHOME|LD_LIBRARY_PATH|LIBRARY_PATH|PATH)$'}
+  BLACKLIST=${3:-'^(GIT_DIR|PYTHONHOME|LD_LIBRARY_PATH|LIBRARY_PATH)$'}
 
   (
     if [ -d "$ENV_DIR" ]; then


joshowen:
+1

jasonpstewart:
More than just a +1... I'm curious as to why the library paths are also blacklisted...  it would be great to make this also work so that if a helper library is installed via the heroku-buildpack-apt package pip could build/link against it. - I've had to do this with swig and libxmlsec1-dev for example - but am using a hacked version of the python buildpack to include the apt libraries and include folders..

stefanosala:
Thanks for the PR @ojacobson 

I've tried your branch, but it doesn't seem to work as expected:
```
➜  receiptful.dev git:(master) staging buildpacks
=== receiptful-staging Buildpack URLs
1. https://github.com/heroku/heroku-buildpack-nodejs
2. https://github.com/ojacobson/heroku-buildpack-python#inherit-path
➜  receiptful.dev git:(master) staging run bash
Running bash on receiptful-staging... up, run.3776
Detected 512 MB available memory, 512 MB limit per process (WEB_MEMORY)
Recommending WEB_CONCURRENCY=1
~ $ node -v
v0.10.40
```

I expected node v4.1.1.

ojacobson:
@jasonpstewart:

> I'm curious as to why the library paths are also blacklisted...

I don't have a great understanding of the history there. Unfortunately, the people who would know are all unavailable right now, so I opted to make a more conservative change here. I'll discuss that with @kennethreitz when he's back, though.

@stefanosala:

Grand. I assume you've rebuilt your app after changing buildpacks? Since you have a ticket open for this already, I can dig into your app and see what I can learn.

ojacobson:
On closer inspection of the code, I think this problem may run deeper. I'll have to look at how we handle `$PATH` at build time more generally, I think.

ojacobson:
Superceded by #253. @stefanosala, if you're interested, I'd love your feedback on that pull request.

kennethreitz:
Looking forward to what you come up with, @ojacobson.

PR #253, opened 2015-10-07
Preserve original build directory (including previously-installed programs) when moving app to /app during build.

ojacobson:
This means that binaries installed in $BUILD_DIR/.heroku/node (let's say) will remain available during build, at the paths previous buildpacks established.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -83,7 +83,7 @@ if [[ ! "$DOCKER_BUILD" ]]; then
 
   # Copy Application code in.
   bpwatch start appdir_stage
-    deep-mv $BUILD_DIR $APP_DIR
+    deep-cp $BUILD_DIR $APP_DIR
   bpwatch stop appdir_stage
 fi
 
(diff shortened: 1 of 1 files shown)


kennethreitz:
does this have any further implications during build time?

stefanosala:
Hi @ojacobson, thanks again for your time.

Unfortunately it doesn't resolve the issue:
```
➜  receiptful.dev git:(hotfix/timeout-script) staging buildpacks
=== receiptful-staging Buildpack URLs
1. https://github.com/heroku/heroku-buildpack-nodejs
2. https://github.com/ojacobson/heroku-buildpack-python#preserve-original-build-dir
➜  receiptful.dev git:(hotfix/timeout-script) staging run bash
Running bash on receiptful-staging... up, run.1582
~ $ node -v
v0.10.40
```

Do you want me to replicate the issue in a small repository?

stefanosala:
Wait, it works for a test app I just created. I'll try again now.

ojacobson:
@kennethreitz:

> does this have any further implications during build time?

I don't think so. I'm delaying the removal of the original build directory (in `/tmp`) until later in the build process, but it's still removed before the buildpack starts copying files back from `/app`. I can't come up with a convincing theory where that would affect the build output, but ideas definitely welcome.

kennethreitz:
Merged :)

PR #270, opened 2016-01-20
Update pip to v8.0.2

edmorley:
Notably this adds support for pip v8's new hash checking mode:
https://pip.pypa.io/en/stable/user_guide/#hash-checking-mode
https://pip.pypa.io/en/stable/reference/pip_install/#hash-checking-mode

For full release notes, see:
https://pip.readthedocs.org/en/stable/news/

NB: Pip 8.x drops support for Python 3.2.


cclauss:
Users of older versions of Python3 will receive this deprecation warning:
__Support for Python 3.0-3.2 has been dropped.__

edmorley:
Those versions are described as unsupported in the readme. I can add a note to the changelog if that helps?

edmorley:
And users of older versions can pin to the last buildpack release?

Or I guess I could add a conditional pip version?

kennethreitz:
Thanks! I'll have to do some testing to ensure that this doesn't break compatibility with my "pip-pop" code, as seen here: https://github.com/heroku/heroku-buildpack-python/tree/master/vendor/pip-pop

edmorley:
I've updated the PR to reflect yesterday's pip 8.0.2 release.

Regarding pip 8.x dropping Python 3.2 support, I've added a note to the changelog.

However if desired I could instead do one of:
1) Add a check for incompatible requested Python version and output a message saying to pin to an older version of the buildpack. (Making the build fail earlier and with instructions).
2) Make the version of pip used conditional (ie for Python 3.2 and equivalent pypi remain on pip 7.1.2)

Let me know your thoughts :-)

erikrose:
@edmorley We should also be sure to check the PGP sig on that, since it's our new trust root. Did you?

edmorley:
I've verified it with the signature at https://pypi.python.org/packages/source/p/pip/pip-8.0.2.tar.gz.asc , however I would hope whomever merges this will not take my word for it and do their due diligence too :-)

Fodder for another ticket, but I guess the buildpack should really also be verifying the PGP signature on the Python download too: 
https://github.com/heroku/heroku-buildpack-python/blob/f172a83d48556b4f0df604760dd6d4f526a591ab/bin/steps/python#L31

edmorley:
This is fixed on master:
https://github.com/heroku/heroku-buildpack-python/commit/b1d1a7cca50868719f3da206f1900856723c023a
https://github.com/heroku/heroku-buildpack-python/commit/8c746e3e177185eadc8174e3cf4dbf0022533fe8
https://github.com/heroku/heroku-buildpack-python/commit/86a53e64799ad0d37ff67c67e7685f0c7a73a738

PR #271, opened 2016-01-27
Add python-ldap support to heroku

josegonzalez:
This allows users to install packages that - for whatever reason - have a dependency upon python-ldap.

It's been tested against an app I use, though honestly we may want to somehow vendor those compiled binaries rather than set them up on the initial push.

Props to @damgad for the [initial work](https://github.com/heroku/heroku-buildpack-python/compare/master...damgad:master). I simply cleaned it up and made it function under the current version of the buildpack.

I think the reasoning behind adding this to the official buildpack is similar to that of GDAL or cryptography; while this is not as common of a requirement as libffi, it is certainly on bar with GDAL, and definitely simplifies the installation of certain software for users on heroku.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -183,6 +183,9 @@ source $BIN_DIR/steps/pip-uninstall
 # Mercurial support.
 source $BIN_DIR/steps/mercurial
 
+# Python-ldap support.
+source $BIN_DIR/steps/python-ldap
+
 # Pylibmc support.
 source $BIN_DIR/steps/pylibmc
 
--- a/bin/steps/python-ldap
+++ b/bin/steps/python-ldap
@@ -20,17 +20,23 @@ if (pip-grep -s requirements.txt python-ldap django-auth-ldap &> /dev/null) then
 
     GROFF_VERSION=${GROFF_VERSION:-1.22.3}
     GROFF_URL=ftp://ftp.gnu.org/gnu/groff/groff-${GROFF_VERSION}.tar.gz
+    GROFF_CACHE_DIR=${CACHE_DIR}/.python-ldap/lib-groff-${GROFF_VERSION}
+    GROFF_BUILD_DIR=${BUILD_DIR}/.python-ldap/lib-groff-${GROFF_VERSION}
 
     BERKLEY_DB_VERSION=${BERKLEY_DB_VERSION:-5.3.28}
     BERKLEY_DB_URL=http://download.oracle.com/berkeley-db/db-${BERKLEY_DB_VERSION}.NC.tar.gz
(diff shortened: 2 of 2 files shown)


kennethreitz:
Thanks for sharing this. Looks like you put a lot of work into it. 

I definitely won't be able to accept it in its current form. If you take a look at the buildpack's support for `cryptography` and friends, the established pattern is to download tarballs of the underlying required libraries/binaries (pre-built) before the pip-install step.

The scripts for generating these tarballs are found in the `builds` directory of this repository.

josegonzalez:
Totally agree, no worries. Definitely knew that would come up, just wanted to be sure this would be accepted if we could overcome that hurdle :)

Should I build the binaries on a heroku box and scarf them down somehow? If so, I'd probably drop support for configuring the library versions. Is that fine by you?

kennethreitz:
If you take a look at the `builds` directory, you'll see a collection of scripts that define how to build libraries that will be pulled in prior to `pip install` (`builds/libraries/vendor`).

If you write the formula for building the library there, then the buildpack can simply download the build artifact and extract it, as the other libraries (e.g. libffi) do.

To build/upload the formula to your own S3 bucket, you can use [bob-builder](https://github.com/kennethreitz/bob-builder).

josegonzalez:
Awesome got it. Will try and revisit this during the weekend.

josegonzalez:
Why are the urls here different:

- https://github.com/heroku/heroku-buildpack-python/blob/master/builds/libraries/libffi
- https://github.com/heroku/heroku-buildpack-python/blob/master/builds/libraries/vendor/libffi

kennethreitz:
@josegonzalez the one under `vendor` is the one you want.

josegonzalez:
Apologies for not getting to this sooner, haven't had time to revisit the original project. This weekend, I promise!

alvarojoao:
@josegonzalez  Thanks!!! It works perfectly!
I was struggling to install the package `django-auth-ldap==1.2.7` in my **Heroku** project.

I just updated my **buildpack** to work with yours:

> `heroku buildpacks:set https://github.com/josegonzalez/heroku-buildpack-python`

and then pushed the code:

> `git push heroku master`

Thanks again

josegonzalez:
I haven't forgotten about this, just haven't had time (just yet) to do the necessary bin changes :(

rinti:
Hey @josegonzalez great work! When trying your solution i get `remote:        AttributeError: 'module' object has no attribute 'OPT_X_TLS_NEVER'` (we use a SSL-connection). Any idea how I could fix this? Happy to hack away on it on a fork, just need a nudge in the right direction.

josegonzalez:
Maybe the required TLS library isn't available? That's judging from this issue: https://mail.python.org/pipermail/python-ldap/2015q1/003504.html

PR #278, opened 2016-02-23
Fix #277 - Install libffi with django[bcrypt] req.

### LINKED ISSUE #277: django[bcrypt] and libffi
[According to Django's docs, the way to include `bcyrpt` is to install `django[bcrypt]`](https://docs.djangoproject.com/en/1.9/topics/auth/passwords/#using-bcrypt-with-django).

Pushing to Heroku with `django[bcrypt]` declared in a `requirements.txt` file fails, as `libffi` is not installed. It appears as though `pip-grep` is not detecting the `bcrypt` requirement.

To be fair, the requirement is not a large one.

        # django/setup.py, lines 25-27, from stable/1.9.x
        extras_require={
            "bcrypt": ["bcrypt"],
        },

The problem disappears if `bcrypt` is defined separately from Django. Nonetheless, it would be nice if the recommended usage in the documentation worked.

I believe the easiest fix would be to add `django[bcrypt]` to the list on line 23 of `heroku-buildpack-python/bin/steps/cryptography`, and will open a PR if that is the desired solution.

jambonrose:
This ensures that libffi, required by bcrypt, is installed when Django
is installed with the bcrypt dependency declared as an extra
(django[bcrypt] or Django[bcrypt]).

### DIFF
--- a/bin/steps/cryptography
+++ b/bin/steps/cryptography
@@ -20,7 +20,7 @@ source $BIN_DIR/utils
 bpwatch start libffi_install
 
 # If a package using cffi exists within requirements, use vendored libffi.
-if (pip-grep -s requirements.txt bcrypt cffi cryptography pyOpenSSL PyOpenSSL requests[security] &> /dev/null) then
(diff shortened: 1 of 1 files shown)


kennethreitz:
:sparkles: :cake: :sparkles:

PR #282, opened 2016-03-01
Verify the checksums of packages downloaded from S3

edmorley:
With pip 8, the integrity of packages downloaded from PyPI during the buildpack compile can be ensured by adding hashes to the requirements files.

However until now there was no guarantee that the Python and/or library packages downloaded from S3 during this buildpack's compile had not been tampered with themselves. Whilst the downloads are performed over HTTPS, also verifying the hash of the downloaded archives protects against:
* S3 bucket compromise
* CA miss-issuance of an s3.amazonaws.com cert
* Protocol or client exploits (Qualys site SSL rating for s3.amazonaws.com is only a 'B')
* Accidental modification of existing files (eg uploading a newer Python version over an older one)

In the case of checksum mismatches, the compile will now be aborted.

For now, if a known checksum is not found for the downloaded file, then only a warning is output. This will prevent breakage if there are any runtimes or stacks uploaded to S3 that are not mentioned in this repo. In the future we may want to make this cause the compile to fail.

The existing checksums were generated via a script (now transposed [below](https://github.com/heroku/heroku-buildpack-python/pull/282#issuecomment-196202677)) run on a cedar-14 dyno and then sorted using `sort -k 2 --version-sort`. There were a number of runtime build files in the vendor directory whose corresponding archive didn't exist on S3 so resulted in an HTTP 403 (https://emorley.pastebin.mozilla.org/8861850).

As part of this PR, the download/extraction steps have been refactored into a separate function in `bin/utils`, since the duplication was becoming excessive.

In addition, the curl download has been made more resilient against failures - it now explicitly fails for non-HTTP 200 responses & also retries. (Inspired by the [implementation in heroku-buildpack-nodejs](https://github.com/heroku/heroku-buildpack-nodejs/blob/v88/lib/binaries.sh#L21).)


edmorley:
**Testing performed:**

All tests were performed using a requirements.txt that triggers the installation of libffi/gdal/libmemcached:
```
cffi
pylibmc
# Newer versions are broken even with the existing buildpack
GDAL==1.9.1
```
...and in-between each test the build cache was cleared using `heroku repo:purge_cache`.

*Variations:*

a) A successful run where checksums were verified ([full log](https://dashboard.heroku.com/apps/python-buildpack-checksum-test/activity/builds/8dae4822-dd0e-45dd-b672-8a6e32411a84) / [diff](https://github.com/heroku/heroku-buildpack-python/compare/master...edmorley:test-checksums)):
```
...
-----> Installing python-2.7.11
-----> Noticed pylibmc. Bootstrapping libmemcached.
-----> Noticed cffi. Bootstrapping libffi.
-----> Noticed GDAL. Bootstrapping gdal.
     $ pip install -r requirements.txt
...
       Successfully installed GDAL-1.9.1 bob-builder-0.0.5 boto-2.39.0 cffi-1.5.2 docopt-0.6.2 envoy-0.0.3 pycparser-2.14 pylibmc-1.5.0
```

b) An expected-fail run with a `runtimes.txt` set to a non-existent runtime, to check we the existing error message is preserved ([full log](https://dashboard.heroku.com/apps/python-buildpack-checksum-test/activity/builds/662409f4-e0ca-4f5c-a497-ec54e23001dd) / [diff](https://github.com/edmorley/heroku-buildpack-python/compare/test-nonexistent-runtime)):
```
...
-----> Installing nonexistent-runtime
 !     Requested runtime (nonexistent-runtime) is not available for this stack (cedar-14).
 !     Aborting.  More info: https://devcenter.heroku.com/articles/python-support
 !     Push rejected, failed to compile Python app
```

c) A successful run where a known checksum wasn't found for the downloaded file, so we output a warning, but don't abort the compile ([full log](https://dashboard.heroku.com/apps/python-buildpack-checksum-test/activity/builds/7a67410d-b422-47f3-8cca-93fce477b9db) / [diff](https://github.com/edmorley/heroku-buildpack-python/compare/test-missing-checksum)):
```
...
-----> Installing python-2.7.11
 !     Skipping verification of python-2.7.11.tar.gz - no known checksum found!
-----> Noticed pylibmc. Bootstrapping libmemcached.
-----> Noticed cffi. Bootstrapping libffi.
-----> Noticed GDAL. Bootstrapping gdal.
     $ pip install -r requirements.txt
       Successfully installed GDAL-1.9.1 bob-builder-0.0.5 boto-2.39.0 cffi-1.5.2 docopt-0.6.2 envoy-0.0.3 pycparser-2.14 pylibmc-1.5.0
```

d) An expected-fail run with a modified known checksums list, that emulates a hash mismatch ([full log](https://dashboard.heroku.com/apps/python-buildpack-checksum-test/activity/builds/5551c983-0307-43c4-be2e-feeef6f2fc52) / [diff](https://github.com/edmorley/heroku-buildpack-python/compare/test-checksum-mismatch)):
```
-----> Fetching set buildpack https://github.com/edmorley/heroku-buildpack-python.git#test-checksum-mismatch... done
-----> Python app detected
 !     Warning: Your application is missing a Procfile. This file tells Heroku how to run your application.
 !     Learn more: https://devcenter.heroku.com/articles/procfile
-----> Installing python-2.7.11
 !     Downloaded file python-2.7.11.tar.gz did not have the expected checksum. Aborting!
 !     Push rejected, failed to compile Python app
```

kennethreitz:
```sh
#!/usr/bin/env bash

# Run this script from a dyno rather than locally, for much faster S3 transfers.
# Copy the output into the relevant checksums file, after sorting with:
# `sort -k 2 --version-sort`

[[ $# -ge 1 ]] || { echo "No stack specified! For example: ./generate-checksums.sh cedar-14"; exit 1; }

STACK=$1
DOWNLOADS_DIR="/app/downloads/$STACK/"

# Runtimes
RUNTIME_URL_PREFIX="https://lang-python.s3.amazonaws.com/$STACK/runtimes/"
RUNTIME_URL_SUFFIX=".tar.gz"

ls -1 /app/builds/runtimes/ \
  | sed -e "s#\(.*\)#$RUNTIME_URL_PREFIX\1$RUNTIME_URL_SUFFIX#" \
  | xargs -n 1 -P 4 wget -nv -N -P $DOWNLOADS_DIR

# Libraries
LIBFFI="https://lang-python.s3.amazonaws.com/$STACK/libraries/vendor/libffi.tar.gz"
GDAL="https://lang-python.s3.amazonaws.com/$STACK/libraries/vendor/gdal.tar.gz"
MEMCACHED="https://lang-python.s3.amazonaws.com/$STACK/libraries/vendor/libmemcache.tar.gz"

wget -nv -N $LIBFFI $GDAL $MEMCACHED -P $DOWNLOADS_DIR

echo -e "\n$STACK checksums...\n"
cd $DOWNLOADS_DIR
sha256sum *
cd -
```

edmorley:
Rebased on master for the recent GDAL changes (and added the checksums for geos and proj).

Meant to say before - after this lands, if the vendored libraries are ever updated, we'll want to add the version number to their filename (rather than overwriting the generic `gdal.tar.gz` etc), to prevent breakage for people who've pinned to a specific version of the buildpack. That said, IMO this is something that should be done regardless of this PR, since otherwise people who believe they've pinned to a deterministic buildpack release can actually have libraries change from under them.

edmorley:
Rebased on master, after the geo-libs script addition.

Updated script to generate the checksums:
```bash
#!/usr/bin/env bash

# Run this script from a dyno rather than locally, for much faster S3 transfers.
# Copy the output into the relevant checksums file, after sorting with:
# `sort -k 2 --version-sort`

[[ $# -ge 1 ]] || { echo "No stack specified! For example: ./generate-checksums.sh cedar-14"; exit 1; }

STACK=$1
DOWNLOADS_DIR="/app/downloads/$STACK/"

# Runtimes
RUNTIME_URL_PREFIX="https://lang-python.s3.amazonaws.com/$STACK/runtimes/"
RUNTIME_URL_SUFFIX=".tar.gz"

ls -1 /app/builds/runtimes/ \
  | sed -e "s#\(.*\)#$RUNTIME_URL_PREFIX\1$RUNTIME_URL_SUFFIX#" \
  | xargs -n 1 -P 4 wget -nv -N -P $DOWNLOADS_DIR

# Libraries
LIBFFI="https://lang-python.s3.amazonaws.com/$STACK/libraries/vendor/libffi.tar.gz"
MEMCACHED="https://lang-python.s3.amazonaws.com/$STACK/libraries/vendor/libmemcache.tar.gz"
GDAL="https://lang-python.s3.amazonaws.com/$STACK/libraries/vendor/gdal.tar.gz"
GEOS="https://lang-python.s3.amazonaws.com/$STACK/libraries/vendor/geos.tar.gz"
PROJ="https://lang-python.s3.amazonaws.com/$STACK/libraries/vendor/proj.tar.gz"

wget -nv -N $LIBFFI $MEMCACHED $GDAL $GEOS $PROJ -P $DOWNLOADS_DIR

echo -e "\n$STACK checksums...\n"
cd $DOWNLOADS_DIR
sha256sum *
cd -
```

kennethreitz:
I don't think I'm going to accept this — for a few reasons I won't get into here. I haven't decided yet though, so I'm leaving the PR open.

edmorley:
Thank you for the update. I don't suppose you could elaborate? It's just it's quite hard to suggest alternative approaches, if I don't know what the issue is :-)

For example instead we could use GPG signing rather than a hardcoded list of checksums (and place the signatures on S3 next to each archive), or if the concern was instead that the list of runtimes in this PR was incomplete (eg due to alternative non-public stacks), then I could make the output silent for cases where we don't have a known checksum etc.

Many thanks!

edmorley:
Rebased on master for the addition of the cedar-14 pypy 5.0.0 and 5.0.1 runtime hashes, as well as the hash update on cedar-14 of python 3.5.1 due to no longer using `--enable-shared` for the compile.

edmorley:
Rebased on master and added hashes for the new pypy 5.1.0 and 5.1.1 runtimes on cedar-14.

@kennethreitz any more thoughts about this PR? :-)

kennethreitz:
I don't think hardcoding hashes is a good idea, for a number of reasons. Mostly, it would pin a given buildpack version (e.g. one older than latest, which is common) to an exact set of binaries that *could* change in the future, for various reasons. Also, a lot of overhead would be involved with maintaining a separate list for each runtime (there will be more in the future). Also, older buildpack users would not be able to use newer security releases of Python, as they can today, being uncoupled.

kennethreitz:
I don't think hardcoding hashes is a good idea, for a number of reasons. 

Mostly, it would pin a given buildpack version (e.g. one older than latest, which is common) to an exact set of binaries that *could* change in the future, for various reasons, and therefore break. Also, a lot of overhead would be involved with maintaining a separate list for each runtime (there will be more in the future). Also, older buildpack users would not be able to use newer security releases of Python, as they can today, being uncoupled. That would be a big step backwards for actual security, not forwards. 

S3 is quite trustworthy, as long as keys are kept safe and rotated appropriately. This seems like a (well-intentioned) unnecessary alignment to cultural best-practices, without a true need or reason beyond "you should do it because you can". If none of the reasons given above were true, and it didn't over-complicate the codebase, it would be a great idea. However, I don't find that to be the case, unfortunately.

If host trust is a major security concern, then TLS is ideally the appropriate layer in which this verification should be occurring.

edmorley:
Thank you for explaining your reasoning :-)

> Mostly, it would pin a given buildpack version (e.g. one older than latest, which is common) to an exact set of binaries that could change in the future, for various reasons, and therefore break.

I think this is where we differ in our outlooks (neither is right or wrong, but means a single approach probably won't work for both) - to me, a particular configuration should be as deterministic as possible - if the underlying vendored archives have changed, they should get new filenames/locations, not be silently overwritten.

> Also, a lot of overhead would be involved with maintaining a separate list for each runtime (there will be more in the future).

This is true (though I'd hoped the provided script would alleviate the concern of this to a certain extent).

> Also, older buildpack users would not be able to use newer security releases of Python, as they can today, being uncoupled. That would be a big step backwards for actual security, not forwards.

In the current PR, if a release is unknown (as would be the case here), it skips the hash check, so this wouldn't be the case.

> If host trust is a major security concern, then TLS is ideally the appropriate layer in which this verification should be occurring.

I agree -- I guess we could always use `--cacert` with curl, which would at least protect against this aspect:

> * CA miss-issuance of an s3.amazonaws.com cert

PR #285, opened 2016-03-11
Update buildpack image to align with marketing

mattgraham:
python


kennethreitz:
Excellent! The old one had code that doesn't run on Heroku anyway — always bothered me :)

:sparkles: :cake: :sparkles:

PR #287, opened 2016-03-21
Keep pre-existing environment variables.

furious-luke:
When using `heroku-buildpack-apt` I found that any values set in `PKG_CONFIG_PATH` was being stomped on by the Python buildpack. I've made a couple of changes to keep the values from previous buildpacks. Is there any reason this is a bad idea? Hopefully not. :)


kennethreitz:
:sparkles: :cake: :sparkles:

PR #290, opened 2016-03-30
Allow users to configure the `/app` directory

josegonzalez:
This is useful for cases in CI where you might want to test in a directory that *isn't* /app.


kennethreitz:
Unfortunately, the Python builds will not operate properly outside of the path provided — so, there's no point in making it configurable at this time.

PR #292, opened 2016-04-06
Allow to add custom env variables before pip-install and collectstatic.

jvenezia:
For exemple we need this when using this buildpack with another buildpack installing packages from conda.

We use it with [jvenezia/heroku-buildpack-python](https://github.com/jvenezia/heroku-buildpack-python) witch installs specific packages with conda. Then the heroku-buildpack-python just have to pickup already installed packages during pip-install.

I'll be happy to discuss this feature and find a better solution. This is the cleanest way i found for my problem at this time.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -52,3 +52,14 @@ Runtime options include:
 - `pypy-5.0.1` (unsupported, experimental)
 
 Other [unsupported runtimes](https://github.com/heroku/heroku-buildpack-python/tree/master/builds/runtimes) are available as well. Use at your own risk. 
+
+
+Add custom environment variables
+--------------------------------
+
+You may need to add custom environment variables during the build. They can be specified with a `.heroku-env` file:
+
+    $ cat .heroku-env
+    PYTHONPATH=/app/.heroku/python/lib/python3.4/site-packages:/app/.heroku/miniconda/envs/condaenv/lib/python3.4/site-packages:$PYTHONPATH
+
+They will be exported before installing packages with pip, and before the Django Collectstatic runner.
--- a/bin/steps/collectstatic
+++ b/bin/steps/collectstatic
@@ -24,6 +24,8 @@ pip-grep -s requirements.txt django Django && DJANGO_INSTALLED=1
 
 bpwatch start collectstatic  # metrics collection
 
+setup-custom-heroku-env
+
 if [ ! "$DISABLE_COLLECTSTATIC" ] && [ -f "$MANAGE_FILE" ] && [ "$DJANGO_INSTALLED" ]; then
     set +e
 
--- a/bin/steps/pip-install
+++ b/bin/steps/pip-install
@@ -1,4 +1,5 @@
 # Install dependencies with Pip.
+setup-custom-heroku-env
 puts-cmd "pip install -r requirements.txt"
 
 [ ! "$FRESH_PYTHON" ] && bpwatch start pip_install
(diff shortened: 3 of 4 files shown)


edmorley:
I think another way that already works without changes to the buildpack, is using the `pre_compile` hook. Unlike some of the other compile steps that are run in a subshell, it is `source`'d be `bin/compile` so can influence the environment for subsequent steps:
https://github.com/heroku/heroku-buildpack-python/blob/0a6ed0de43c549508d3b98ac740b86ed185ccf89/bin/compile#L130

Perhaps try adding the environment variable exports to a file in your app's repository named `bin/pre_compile` and see how you get on? :-)

jvenezia:
This may be a nicer solution :)

Unfortunately it is not working for me right now:

I added the precompile hook `bin/pre_compile`:
```
$ cat bin/pre_compile
export PYTHONPATH=/app/.heroku/python/lib/python3.4/site-packages:/app/.heroku/miniconda/envs/condaenv/lib/python3.4/site-packages:$PYTHONPATH
```

The hook is running during the build as a can see this message: `-----> Running pre-compile hook`.

But it seems to be ignored during pip-install, pip ignores my new PYTHONPATH and tries to install packages that I already installed with conda (numpy, scipy, scikit-learn).

Any Idea ?

jvenezia:
Okay the problem is in the `sub-env` function witch is reseting `PYTHONPATH`:
https://github.com/heroku/heroku-buildpack-python/blob/master/bin/utils#L93

And it is called after executing the hooks:
https://github.com/heroku/heroku-buildpack-python/blob/master/bin/steps/hooks/pre_compile#L6

Also called before pip-install:
https://github.com/heroku/heroku-buildpack-python/blob/master/bin/compile#L205

And collectstatic:
https://github.com/heroku/heroku-buildpack-python/blob/master/bin/compile#L214

So my `PYTHONPATH` definition is ignored...

Is it possible to make `sub-env` less agressive? This looks like a issue to me. What do you think?

kennethreitz:
There are many ways to approach this, but I'm going to let you know that a `.heroku-env` file is absolutely not going to be accepted.

jvenezia:
Haha ^^
Okay, it would be nice to have a little bit more details :) 

I'll be happy to help. What approaches are you considering ? I like the hook solution. Making the `sub-env` less agressive seems ok to me.

jasonpstewart:
Hi Kenneth,
We're looking for a solution to this as well.  (We use the apt buildpack and install swig and lib-xmlsec1 and need to adjust the build env vars so that pip can install dm.xmlsec.binding and M2Crypto).  

Any suggestions you have on how to proceed without needing a forked/hacked buildpack I think would be welcomed by a good number of people.

kennethreitz:
Well, `PYTHONHOME` and `PYTHONPATH` are sacred territory in the eyes of the Python buildpack, and no customization is currently possible, or will be. Other variables, however, should be available.

jvenezia:
Well again, it would be great to have more details on this.

Okay so no `.heroku-env` and no customisation of `PYTHONPATH` ever.

What do you suggest ? What are your ideas to achieve this ? What are the "many approaches" ? What other variables could be available and how will this solve the problem ?

kennethreitz:
Well this buildpack and the Python buildpack are not currently compatible at this time. It seems like that is what you're trying to accomplish. This has been considered, and may be revisited in the future, but for the time being, this concept is not being pursued.

jvenezia:
Okay thanks for the explanation.

It's not only for your conda buildpack, but any buildpack that installs packages before the python buildpack.

PR #297, opened 2016-05-10
remove Pylibmc reference in cryptography comment

whit537:


kennethreitz:
Either is fine. Technically, the file should be rennamed and the comment should remain the same.

kennethreitz:
OH it says "Pylibmc" :P

whit537:
Yeah, presumably it was an artifact of copying/pasting/editing [`bin/steps/pylibmc`](https://github.com/whit537/heroku-buildpack-python/blob/e4aa40216bd55197040ba2fa4729c7289f09add0/bin/steps/pylibmc).

👍

kennethreitz:
@whit537 ❤️ <-- die for penny-defacer

PR #298, opened 2016-05-10
clean up pip-grep in the `not silent` case

whit537:
The extraneous `format` call is harmless in itself, but it's possible for `requirement.req` to be `None` here, which results in `AttributeError: 'NoneType' object has no attribute 'project_name'`.


kennethreitz:
:sparkles: :cake: :sparkles:

PR #301, opened 2016-05-13
Added another package that relies on cffi

alex:


kennethreitz:
Thanks!

This will get deployed whenever the next buildpack release occurs.

alex:
👍  in the meantime I just list `cffi` in my `requirements.txt` explicitly

PR #303, opened 2016-05-18
Update pip-install

ettoreleandrotognoli:


kennethreitz:
This is an interesting option, and has been considered before. You're the first person to actually send a pull request.

I feel like it's likely not the best idea, given that it would be a big change in expected behavior for existing users/applications. But, if given enough thought, it might be fine. I need to think about it. 

Learning towards no, though.

PR #306, opened 2016-06-01
fix(export-env): export app config vars

RenaultAI:
Explicitly call the sub-env function to export an app's config variables, which are in the ENV_DIR directory.

Also ensure the environment variables are exported in the main shell of the program instead of a subshell. The use case here is to pass flags to pip via PIP_* environment variables. https://pip.pypa.io/en/stable/user_guide/#environment-variables

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -183,6 +183,9 @@ set -e
 
 mkdir -p $(dirname $PROFILE_PATH)
 
+# Export application's configuration variables
+sub-env $ENV_DIR
+
 # Install Python.
 source $BIN_DIR/steps/python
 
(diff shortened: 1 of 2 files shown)


kennethreitz:
these environment variables are only to be made available to specific steps of the build process, not the entire compile phase. They are sensitive. For example, `$ pip install` has the ability to run any arbitrary Python code.

kennethreitz:
@RenaultAI what flags specifically would you want to use? No one has requested this before :)

RenaultAI:
You're right! I neglected to see the change from a security point of view.

The specific pip flag is --process-dependency-links, which can be enabled via the PIP_PROCESS_DEPENDENCY_LINKS environment variable. With pip version 8.0.0 and newer, this flag can be provided in the requirements file however. I created the pull request to provide a consistent solution.

Perhaps a better change is to add ^PIP_ to the environment variable whitelist?

kennethreitz:
The intended (and recommended) solution is to add `--process-dependency-links` to your `requirements.txt` file.

PR #319, opened 2016-08-02
Fixes to the library vendoring's 'already exists' checks

edmorley:
* Makes libffi vendoring check for the existing directory in the correct location, to avoid constant re-downloading for every compile. (For both the cedar and cedar-14 stacks, the libffi directory is named `libffi-3.1` not `libffi-3.1.1`.)
* Makes the geo-libs vendoring step work even if someone previously had `GDAL` in their requirements file, so already had `gdalserver` vendored by `bin/steps/gdal`.
* Avoids repetition of vendor library environment variable setting.


kennethreitz:
These changes look like solid improvements, thank you. I will have to test them before pushing to production.

PR #329, opened 2016-09-16
Add argon2 to crypotgraphy step

jambonrose:
- Add argon2-cffi
- Add django[argon2] option from Django docs:
  https://docs.djangoproject.com/en/1.10/topics/auth/passwords/#argon2-usage


kennethreitz:
✨🍰✨

PR #340, opened 2016-12-16
Greatly simplify code copying

dzuelke:
A few simple Symlinks so Python and modules can find themselves in `/app/.heroku/…` are enough :)


kennethreitz:
this should fix the exports too

kennethreitz:
david will update pull request with final fix for cache bug

dzuelke:
Confirmed working even with native packages (`psycopg2`).

dzuelke:
Confirmed fixes Python breakage for subsequent buildpacks.

dzuelke:
Caching is still broken right now so only the first push succeeds. Will fix.

kennethreitz:
✨🍰✨

ojacobson:
👍 huge improvement

PR #341, opened 2016-12-17
Docker for heroku-16

kennethreitz:


kennethreitz:
will be merging next when next python is released

PR #354, opened 2016-12-30
Update README.md to current versions of Python 2, Python 3, Pypy

matsulib:
For consistency with [Heroku Python Support](https://devcenter.heroku.com/articles/python-support).
In addition, Requests was also updated.


kennethreitz:
✨🍰✨

PR #358, opened 2017-01-28
PROPOSAL - call Django "manage.py migrate" in compile process

drgarcia1986:
I think it's useful for deploy pipeline with migrations (instead of run `migrate` command after deploy with `heroku run`  command)

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -200,6 +200,9 @@ deep-cp /app/.heroku/src $BUILD_DIR/.heroku/src
 # Django collectstatic support.
 sub-env $BIN_DIR/steps/collectstatic
 
+# Django migrate support.
+sub-env $BIN_DIR/steps/django-migrate
+
 # Create .profile script for application runtime environment variables.
 set-env PATH '$HOME/.heroku/python/bin:$PATH'
 set-env PYTHONUNBUFFERED true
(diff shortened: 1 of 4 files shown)


edmorley:
Running tasks that affect external resources (eg a DB migrate) during slug creation is not advisable, for the following reasons:
* The slug compile only occurs once, so these tasks won't be re-run in the case of [rolling back a release](https://devcenter.heroku.com/articles/releases#rollback) or when using [pipelines promotion](https://devcenter.heroku.com/articles/pipelines#promoting).
* Slug compilation time is limited to 15 minutes, and a DB migration can often take longer than this.

Heroku has a relatively new feature, "release phase" that is designed specifically for running such tasks:
https://devcenter.heroku.com/articles/release-phase

This can also be used for much more than just running migrate, eg:
* reporting New Relic deploys
* running Django system checks with the `--deploy` option (which checks secure settings are being used etc)
* importing Django fixtures

...and the tasks are allowed to run for up to 24 hours.

For an example, see:
https://github.com/mozilla/treeherder/blob/b14eb95c571e00683d5fe9a283ac532623c7f23f/Procfile#L15
https://github.com/mozilla/treeherder/blob/b14eb95c571e00683d5fe9a283ac532623c7f23f/bin/pre_deploy

drgarcia1986:
make sense, thanks @edmorley

kennethreitz:
Thanks @edmorley

PR #363, opened 2017-03-03
Add "PIP_" prefix environment variables to pip-install

williammcbroom:
This change allows user to add options to `pip` through the Config Variables instead of just the `requirements.txt` file. For example a user can add the variable `PIP_EXTRA_INDEX_URL` which may contain a username and password which is better suited as a Config Variable instead of having to be committed in to the repository code by adding it to the `requirements.txt` file.


williammcbroom:
@kennethreitz Closed without even a comment? I feel like this could be a very useful feature for others. Why do you think it shouldn't be merged?

kennethreitz:
interesting, github won't let me reopen this PR

williammcbroom:
@kennethreitz Weird. Maybe it's been too long? I'm okay with the changes in PR #639. As long as we can get `PIP_EXTRA_INDEX_URL` set then i'm good.

PR #366, opened 2017-03-14
Ortools

christianklotz:


edmorley:
Note: or-tools could now ship wheels for Linux (thanks to [PEP 513](https://www.python.org/dev/peps/pep-0513/)), which at an albeit brief glance at google/or-tools#197 looks like it would solve the pip incompatibility?

christianklotz:
@edmorley thanks for pointing that out 🙇. I hope or-tools will receive the required update soon.

PR #371, opened 2017-04-04
don't unset DYNO (we're running on Heroku)

### LINKED ISSUE #280: Should environment variables present in a running dyno be unset in the build environment?
As a means of detecting if a Django application is running on a Heroku dyno we had been checking in the environment for the presence of the DYNO variable.

`if os.environ.get('DYNO'):
    do_stuff()
`

This variable is set in a running dyno, however since 669f0b0 it is no longer set during the build process.
This resulted in a difficult to troubleshoot issue for us where the Django staticfiles storage backend was different on the running dyno versus when collectstatic is being run (versioned media files were not being copied to the STATIC_ROOT).

The [$DYNO variable is documented as "experimental and subject to change or removal"](https://devcenter.heroku.com/articles/dynos#local-environment-variables) so I can understand if it is removed, however it might be beneficial if the environment is consistent between the build environment and the running dyno to avoid similar issues to this.

We've already moved away from this type of Heroku-specifc configuration, but I figured I would raise the issue here as it has the potential to cause confusion for others in the future. Thanks!

kennethreitz:
Fixes #280

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -48,7 +48,7 @@ export PATH=$PATH:$ROOT_DIR/vendor/:$ROOT_DIR/vendor/pip-pop
 
 # Sanitizing environment variables.
 unset GIT_DIR PYTHONHOME PYTHONPATH
-unset RECEIVE_DATA RUN_KEY BUILD_INFO DEPLOY LOG_TOKEN DYNO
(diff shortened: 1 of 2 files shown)


kennethreitz:
@dzuelke STACK is now blacklisted.

PR #321, opened 2016-08-05
Restore the build cache prior to running bin/pre_compile

edmorley:
So that any changes made to `.heroku/` within pre_compile (such as installing additional libraries required for the later pip install) are not clobbered by the cache being copied on top of it afterwards.

Fixes #320.

PR #372, opened 2017-04-04
Restore the build cache prior to running bin/pre_compile

### LINKED ISSUE #320: Run `bin/pre_compile` after the cache is restored, not before
The `bin/pre_compile` script is currently run not just before the main compile steps (eg installing Python/pip install etc), but also before the cache restoration step. 

This means that if a `bin/pre_compile` script modifies eg `/app/.heroku/vendor` (such as extracting a required library that will be required later during pip install), then:
(a) any changes to `/app/.heroku/vendor` will be clobbered by the cache being copied over it (if the files already existed in the cache) - which is a problem if the library has been updated to a newer version (eg imagine libmysqlclient 5.7.13 is cached and pre_compile just installed the newer 5.7.14 into `/app/.heroku/vendor`, the older cached version will overwrite it).
(b) any "already exists so can skip library download/extraction" check in `bin/pre_compile` has to remember to check `$CACHE_DIR/.heroku/vendor` not `/app/.heroku/vendor` to see whether it can skip downloading/extracting the library.

Whilst the logic in `bin/pre_compile` can be tweaked to instead operate against `$CACHE_DIR` to work around the above (ie extract the files there and then rely on them being moved to `/app/.heroku/vendor` later), this means using completely different paths in `pre_compile` compared to `post_compile` and thus having to annotate with a number of comments explaining why.

I think it would be more intuitive if the cache restore step was run prior to `bin/pre_compile`, not after.

I'll open a PR shortly :-)

edmorley:
So that any changes made to `.heroku/` within pre_compile (such as installing additional libraries required for the later pip install) are not clobbered by the cache being copied over afterwards.

Fixes #320.

This is a clone of #321 since GitHub doesn't allow PRs to be reopened in some cases (think if they are rebased whilst closed?).


kennethreitz:
Requested a review by @dz. @edmorley has been suggesting this change for 6 months. It looks good to me, and CI passes, so I don't think any harm is being done.

kennethreitz:
Once review is complete, I'll test manually before merging.

edmorley:
Many thanks :-)

kennethreitz:
✨🍰✨

kennethreitz:
Sorry it took so long!

PR #374, opened 2017-04-25
basics of metrics

kennethreitz:


kennethreitz:
work in progress

kennethreitz:
This also adds a new warnings check for inclusion of the six library, which is needed with the release of the latest setuptools chain, which has a few dependencies, unfortunately. 

Plan to add the other dependencies as well, eventually, but six is the one people run into the most. 

I want to look into removing these entries from `requirements.txt` automatically.

bin/utils
```diff
@@ -1,5 +1,12 @@
+#!/usr/bin/env bash
 shopt -s extglob
 
+# The standard library.
+if [[ ! -f  /tmp/stdlib.sh ]]; then
+  curl -s https://raw.githubusercontent.com/heroku/buildpack-stdlib/v2/stdlib.sh > /tmp/stdlib.sh
```
jmorrell: Add retry flags `--retry 5 --retry-max-time 15`

bin/utils
```diff
@@ -108,3 +74,8 @@ sub-env() {
   )
 }
 
+# Measure the size of the Python installation.
+measure-size() {
```
jmorrell: The name is pretty generic for being so specific. `measure-python-directory`?

bin/warnings
```diff
@@ -37,6 +40,18 @@ distribute-included() {
     puts-warn "This library is automatically installed by Heroku and shouldn't be in"
     puts-warn "Your requirements.txt file. This can cause unexpected behavior."
     puts-warn "  -- Much Love, Heroku."
+    mcount 'warnings.distribute'
+  fi
+}
+
+six-included() {
```
jmorrell: This should probably be a different PR, but we can roll with it

bin/utils
kennethreitz: it's the exact same name as the one used in the node buildpack

jmorrell: Then I'll update that one :)

bin/compile
```diff
@@ -137,10 +141,14 @@ if [[ $BUILD_DIR != '/app' ]]; then
 fi
 
 # Install Python.
+let start=$(nowms)
```
kennethreitz: this is the documented way of doing this.

bin/utils
kennethreitz: good call

PR #386, opened 2017-05-18
Update pipenv to remove broken option.

jmduke:
In pipenv 4.0.0, `--no-dashes` is removed and causes the command to fail:

```
$ pipenv lock --requirements --no-hashes
Error: no such option: --no-hashes
```

As far as I can tell, this breaks all heroku deployments that use pipenv since there's no way to specify a version of pipenv.  Removing the flag will presumably fix this.

### DIFF
--- a/bin/steps/pipenv
+++ b/bin/steps/pipenv
@@ -5,6 +5,6 @@ if [[ -f Pipfile ]]; then
 
         /app/.heroku/python/bin/pip install pipenv --upgrade &> /dev/null
 
-        /app/.heroku/python/bin/pipenv lock --requirements --no-hashes > $BUILD_DIR/requirements.txt 2> /dev/null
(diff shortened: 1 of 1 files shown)


jmduke:
(It looks like #385 obviates this change, but if that's a WIP then it might be worth getting this out the door ahead of it?  If there's another way to remediate this issue that I'm missing, that'd be good too!)

jacebrowning:
@kennethreitz I really need either this change and/or #385. All my Heroku deployments using `pipenv` are broken.

jacebrowning:
Until then, anyone can add this to `app.json` if they need it:

```
"buildpacks": [
  {
    "url": "https://github.com/jmduke/heroku-buildpack-python.git#3d09fbaccf3d78bef76f09634655ddc01c0bca46"
  }
],
```

edmorley:
This has been fixed by the refactor in #385 :-)

kennethreitz:
thanks for the contribution!

Issue #379, opened 2017-05-09
Heroku-16 Python runtime archives contain additional packages (boto/bob/...)


edmorley:
The cedar-14 Python runtime archive does not contain any already installed packages:
```bash
$ curl -sSf https://lang-python.s3.amazonaws.com/cedar-14/runtimes/python-2.7.13.tar.gz  | tar -zt  --wildcards '*/site-packages/*.py'
tar: */site-packages/*.py: Not found in archive
tar: Exiting with failure status due to previous errors
```

However the Heroku-16 equivalent does:
```
$ curl -sSf https://lang-python.s3.amazonaws.com/heroku-16/runtimes/python-2.7.13.tar.gz | tar -zt  --wildcards '*/site-packages/*.py'
lib/python2.7/site-packages/docopt.py
lib/python2.7/site-packages/boto/storage_uri.py
lib/python2.7/site-packages/boto/requestlog.py
lib/python2.7/site-packages/boto/auth_handler.py
...
lib/python2.7/site-packages/bob/cli.py
lib/python2.7/site-packages/bob/utils.py
lib/python2.7/site-packages/bob/models.py
...
lib/python2.7/site-packages/pip-8.1.2-py2.7.egg/pip/cmdoptions.py
lib/python2.7/site-packages/pip-8.1.2-py2.7.egg/pip/wheel.py
lib/python2.7/site-packages/pip-8.1.2-py2.7.egg/pip/status_codes.py
...
lib/python2.7/site-packages/envoy/core.py
lib/python2.7/site-packages/envoy/__init__.py
```

There are appear to be other files present that were not in cedar-14, eg:
```
requirements-declared.txt
requirements-installed.txt
requirements-stale.txt
```

I'm guessing the method used to build the archives has accidentally bundled parts of this buildpack's development environment into the archive?

edmorley:
There also appear to be some 2.7.12 related files in the 2.7.13 Heroku-16 archive, eg:
```
lib/python2.7/lib-dynload/Python-2.7.12-py2.7.egg-info
lib/python2.7/lib2to3/Grammar2.7.12.final.0.pickle
lib/python2.7/ensurepip/_bundled/pip-8.1.1-py2.py3-none-any.whl
```

I'm presuming the builds were generated sequentially in the same environment, and perhaps these are leftover files from the previous build?

kennethreitz:
[...]

PR #387, opened 2017-05-25
Improve the workflow for generating binaries for Heroku-16

edmorley:
* Updates bob-builder to a version that clears out the previous build directory, which prevents the stale files seen in #379.
* Adds a `buildenv-heroku-16` Make command to simplify the building and use of the Heroku-16 binary build environment.
* Sets `S3_BUCKET` and `S3_PREFIX` in the Docker image, to save having to do so manually each time.
* Removes the duplication of the bob-builder dependency between `Dockerfile` and `requirements.txt`.
* Disables the pip version check during Docker build, to avoid the noisy stderr warning about Ubuntu 16.04 pip being older (v8.1.1).
* Adds a `.dockerignore` to speed up the Docker build, by reducing the build context transferred to the daemon from 60MB to 2MB.
* Applies some Dockerfile best practices like disabling/removing cached files (since layer invalidation makes caching pointless).


kennethreitz:
will merge once tests pass!

kennethreitz:
waiting for CI to pass

kennethreitz:
✨🍰✨

PR #395, opened 2017-05-30
use pkg_resources to check for distributions

### LINKED ISSUE #359: collectstatic does not run when using setup.py
I am using `setup.py` instead of `requirements.txt` to specify my dependencies. The dependencies are installed correctly; however, the `collectstatic` step does not detect that Django is installed.

I verified this by adding `pip-grep requirements.txt django Django` to `bin/post_compile`. Since `requirements.txt` is generated and only contains `-e .` when using `setup.py`, `pip-grep` outputs `Not found.`.

This probably affects any other steps that rely on `pip-grep` to check dependencies.

davidism:
This changes the implementation of `pip-grep` to use `pkg_resources.get_distribution` instead of parsing `requirements.txt` manually. This makes the buildpack more accurate for some project configurations.

This change is backwards compatible, scripts calling this with a requirements file will continue to work, since "requirements.txt" is not a package.

fixes #359

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Python Buildpack Changelog
 
+# 104
+
+unreleased
+
+- Use `pkg_resources` to check if a distribution is installed instead of
+  parsing `requirements.txt`. ([#395][395])
+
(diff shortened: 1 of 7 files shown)


kennethreitz:
✨🍰✨

Issue #393, opened 2017-05-26
The pip-uninstall step is always a no-op


edmorley:
The pip-uninstall step aims to uninstall any stale packages from the buildpack python site-packages directory:
https://github.com/heroku/heroku-buildpack-python/blob/master/bin/steps/pip-uninstall

Unfortunately at the moment this isn't working since the sequence of steps are:
* Run `bin/steps/pip-install`, which:
  - runs pip install
  - makes a copy of the app's requirements.txt at `.heroku/python/requirements-declared.txt`
  - saves the output from `pip freeze` to `.heroku/python/requirements-installed.txt`
* Run `bin/steps/pip-uninstall`, which:
  - uses pip-diff to compare `.heroku/python/requirements-declared.txt` against `requirements.txt`, and generate a list of stale requirements
  - passes that list to `pip uninstall`

...which means that `requirements-declared.txt` used by the uninstall step is created by that run's install step (rather than the previous build), so is always identical to `requirements.txt`, so no packages are ever uninstalled.

This worked previously, since the pip-uninstall script was run prior pip-install - but the order was reversed in:
https://github.com/heroku/heroku-buildpack-python/commit/f27a84e0155c7752630aae6c93e230d2a24a880a

I also spotted that:
* `requirements-installed.txt` (the output from `pip freeze`) is never used
* even with the above fixed, the uninstall process wouldn't handle apps that have includes in `requirements.txt` (eg `-r requirements/foo.txt`). 
  - One way to fix this is to try and flattern the requirements chain using the pip API, like pip-diff already does, or else to make use of the sync command from [pip-tools](https://github.com/jazzband/pip-tools/blob/master/piptools/sync.py).
  - Though understand this might be a "wontfix and add to docs" case, since apps are encouraged to use a single requirements file at the root of the repo.

I'll take more of a look at this when I'm back next week :-)

kennethreitz:
Thanks for reporting this!

kennethreitz:
@edmorley I changed the ordering here again.

edmorley:
> @edmorley I changed the ordering here again.

Indeed! Fixed by:
https://github.com/heroku/heroku-buildpack-python/pull/397/files#diff-c7b43338c09c3c9afd420fd8b6bb5198R174

I've split out the remaining parts of this to #411 and #412.

Many thanks!

edmorley:
> Indeed! Fixed by:
[...]

PR #397, opened 2017-05-30
Don't install packages that could mess packaging up

kennethreitz:


vendor/pip-pop/pip-clean
```diff
@@ -0,0 +1,40 @@
...
+  -h --help     Show this screen.
+"""
+from docopt import docopt
+
+BAD_PACKAGES = ['appdirs', 'packaging', 'pyparsing', 'six', 'setuptools', 'distribute']
+
+def main():
+    args = docopt(__doc__, version='pip-grep')
+    req_file = args['<req-file>']
+
```
cclauss: Could this be streamlined?

```python
def good_package(line):
	package_name = line.split('=')[0].split('<')[0].split('>')[0]
	return package_name not in BAD_PACKAGES

lines = [line for line in f if good_package(line)]
```

kennethreitz: I want to do as little parsing as possible.

cclauss: Understood...  see other package names that begin with but do not end with `six` for instance https://pypi.python.org/pypi?%3Aaction=search&term=Six&submit=search

bin/steps/setuptools
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+# Syntax sugar.
+source $BIN_DIR/utils
+
+pip-clean requirements.txt
```
hone: Should we have a message that tells people we're replacing stuff in their `requirements.txt` with things in the Python buildpack or overwriting their `requirements.txt` file at all.

Also, how does this affect pipenv?

vendor/pip-pop/pip-clean
kennethreitz: I was planning to cross that bridge when I came to it, which I honestly doubt would happen any time soon. I'll see what I can do though.

hone:
Should we have a message that tells people we're replacing stuff in their requirements.txt with things in the Python buildpack or overwriting their requirements.txt file at all.

Also, how does this affect pipenv?

kennethreitz:
Pipenv has this functionality built-in. 

I don't think a message is needed, but I could add one saying "Removing packaging libraries from requirements.txt".

kennethreitz:
In my mind, if someone has any of these listed in their requirements.txt, it is accidental, not intentional.

kennethreitz:
e.g. they don't understand what's going on. so, i don't want to overload them with information.

hone:
@kennethreitz is there a place you could link them to that would help?

kennethreitz:
No.

kennethreitz:
This is a new problem, and one that is temporary and will go away soon (but we need to have this code for people using this version range of setuptools locally for the future). 

It's not well-understood.

vendor/pip-pop/pip-clean
kennethreitz: @cclauss i updated the code; integrated what you suggested and made a few improvements

kennethreitz:
Waiting, waiting, waiting... for CI to pass.

bin/steps/setuptools
jayvdb: missing EOL

vendor/pip-pop/pip-clean
```diff
@@ -0,0 +1,33 @@
...
+BAD_PACKAGES = ['appdirs', 'packaging', 'pyparsing', 'six', 'setuptools', 'distribute']
+
+
+def good_package(line):
+    package_name = line.split('=')[0].split('<')[0].split('>')[0].split(' ')[0].split('#')[0].split('\n')[0]
+    return package_name not in BAD_PACKAGES
+
+def main():
+    args = docopt(__doc__, version='pip-grep')
```
jayvdb: `version='pip-grep'`

kennethreitz: thanks!

bin/steps/setuptools
kennethreitz: thanks!

jayvdb:
> In my mind, if someone has any of these listed in their requirements.txt, it is accidental, not intentional.
> e.g. they don't understand what's going on. so, i don't want to overload them with information.

The more usual explanation is they have put them in because they have minimum and maximum requirements.
`setuptools` installs `pkg_resources` which is the main mechanism people use for fetching plugins (implicit namespaces is the newer alternative).
And more importantly, because of the fact that pip doesnt have a proper resolver, sometimes it is *necessary* to mention one of these packages in `requirements.txt` so that pip follows a particular path in its resolution strategy.  You remove one of them of `requirements.txt`, and pip behaves differently when resolving indirect dependencies many levels deep.
At the very least, there should be a note to say why the deploy is using different versions from the versions they see when they deploy the app somewhere else.

vendor/pip-pop/requirements.txt
```diff
@@ -0,0 +1,17 @@
...
+python-dateutil==2.6.0
+pytz==2017.2
+pytzdata==2017.2
+regex==2017.5.26
+requests==2.17.3
+ruamel.ordereddict==0.4.9
+ruamel.yaml==0.14.12
+tzlocal==1.4
+urllib3==1.21.1
+
```
jayvdb: missing EOL.

kennethreitz:
Related: https://github.com/heroku/heroku-buildpack-python/issues/393

vendor/pip-pop/pip-clean
```diff
@@ -0,0 +1,33 @@
...
+Options:
+  -h --help     Show this screen.
+"""
+from docopt import docopt
+
+BAD_PACKAGES = ['appdirs', 'packaging', 'pyparsing', 'six', 'setuptools', 'distribute']
+
+
+def good_package(line):
+    package_name = line.split('=')[0].split('<')[0].split('>')[0].split(' ')[0].split('#')[0].split('\n')[0]
```
kennethreitz: '\n' is included in each line

jayvdb: Buggy. c.f. https://github.com/heroku/heroku-buildpack-python/issues/400

edmorley:
> This is a new problem, and one that is temporary and will go away soon

Was this to work around setuptools unvendoring their dependencies in v34.0.0 (which they've then reverted in v36.0.0) and relying on `install_requires` instead? (Which presumably could cause issues if the versions they required conflicted with those in requirements.txt?)

kennethreitz:
@edmorley yes

PR #404, opened 2017-06-03
Revert "Don't install packages that could mess packaging up"

kennethreitz:
Reverts heroku/heroku-buildpack-python#397


kennethreitz:
setuptools (thankfully) made some changes that makes this no longer necessary! \o.

PR #417, opened 2017-06-29
Allow environment variables during pip install, fixes #416

### LINKED ISSUE #416: Please enable 'sub-env' at the pip install step
This is basically to support installing from private repos without having to expose credentials.

I want to add an API token for GitLab or GitHub into my heroku dashboard and have that used to install a private repo, like so:

```
setup(
    install_requires=[
        'my_private_library'
    ],
    dependency_links=[
        'git+https://%s@gitlab.com/private/private.git#egg=my_private_library-1.2.3'%os.environ.get('GITLAB_TOKEN'), 
    ]
)
```

With a `requirements.txt` like this:
```
--process-dependency-links
-e .
```

Everything works fine locally, if I do: `GITLAB_TOKEN=my_secret_key pip install -r requirements.txt`

But on heroku, the environment variable isn't available.

I'll submit a PR soon, but am open to feedback.

LegoStormtroopr:


kennethreitz:
Well, these tests don't pass.

LegoStormtroopr:
Hmmm... I didn't think it'd fail that dramatically - I'm happy to put some work into fixing this if its of interest.

LegoStormtroopr:
Ok, this isn't as simple as I'd hoped. Gimme some time to figure this out.

williammcbroom:
I am very interested in pr and hope that you are able to get it fixed and approved.

edmorley:
Ah the reason the tests are failing is that `sub-env` only uses the first argument (`$1`) and discards the rest, so every pip invocation is calling just `/app/.heroku/python/bin/pip` without the `install` and following arguments.

The fix is to modify `sub-env` to use `"$@"` instead (note the surrounding double quotes, to avoid the problems mentioned [here](https://stackoverflow.com/a/4824637)).

I've checked the existing consumers of `sub-env` and all of them only pass one argument, so handily won't be affected.

LegoStormtroopr:
@edmorley thanks for that help, I'll have another go and see what I can do :)

kennethreitz:
@edmorley want to get a PR open to the buildpack stdlib fixing this?

edmorley:
> @edmorley want to get a PR open to the buildpack stdlib fixing this?

Sure - though this buildpack defines it's own sub-env, so the next step after fixing buildpack-stdlib will be switching away from that :-)

p-himik:
I guess there's no need to create a separate feature request for `PIP_*` environment variables (https://pip.pypa.io/en/stable/user_guide/#environment-variables) - they will be taken into account automatically after this PR is fixed and merged. Is that correct?

edmorley:
@p-himik - that's correct :-)

edmorley:
> > @edmorley want to get a PR open to the buildpack stdlib fixing this?
> 
> Sure - though this buildpack defines it's own sub-env, so the next step after fixing buildpack-stdlib will be switching away from the one here :-)

I've filed heroku/buildpack-stdlib#37 - once the desired approach is clearing, I'll open a PR for it.

LegoStormtroopr:
I finally got around to fixing this properly - happy for feedback :)

kennethreitz:
Removed this. Caused all kinds of problems.

edmorley:
@LegoStormtroopr - some extra context:
* older stacks didn't used to have the `init.d/` scripts, so buildpacks used to have to set certain environment variables using the `bin/release` script (such as `PYTHONHOME` and `PYTHONPATH`)
* apps that were created back then still have these environment variables set, which is problematic when they are allowed to pass through to the pip invocation
* whilst these unused and out of date environment variables could be unset from apps by their owners, it causes breakage until people do so, resulting in a lot of support tickets - which is why Kenneth reverted this in v115

For your use-case I'd recommend trying the alternative approaches in https://github.com/heroku/heroku-buildpack-python/issues/416#issuecomment-327206137 - and in the meantime pinning to v113 which includes this PR.

LegoStormtroopr:
No worries, I'll stick to the work arounds for now. Thanks again.

PR #438, opened 2017-08-03
Shellcheck Compliance

kennethreitz:
## TODO

- [x] `compile`
- [x] `release`
- [x] `detect`
- [x] `warnings`
- [x] `utils`
- [x] `test-compile`
- [x] `steps/collectstatic`
- [x] `steps/cryptography`
- [x] `steps/eggpath-fix`
- [x] `steps/eggpath-fix2`
- [x] `steps/gdal`
- [x] `steps/geo-libs`
- [x] `steps/hooks`
- [x] `steps/mercurial`
- [x] `steps/nltk`
- [x] `steps/pip-install`
- [x] `steps/pip-uninstall`
- [x] `steps/pipenv`
- [x] `steps/pipenv-python-version`
- [x] `steps/pylibmc`
- [x] `steps/python`

---------------------

- [x] add shellcheck to test script


kennethreitz:
this already uncovered some great bugs!

kennethreitz:
tests passing, ✨🍰✨

bin/compile
```diff
@@ -92,148 +98,170 @@ if [[ ! -f Procfile ]]; then
...
+cp -R "$CACHE_DIR/.heroku/python-stack" .heroku/ &> /dev/null || true
+cp -R "$CACHE_DIR/.heroku/python-version" .heroku/ &> /dev/null || true
+cp -R "$CACHE_DIR/.heroku/vendor" .heroku/ &> /dev/null || true
+if [[ -d "$CACHE_DIR/.heroku/src" ]]; then
+  cp -R "$CACHE_DIR/.heroku/src" .heroku/ &> /dev/null || true
 fi
 
 # Experimental pre_compile hook.
-source $BIN_DIR/steps/hooks/pre_compile
+# shellcheck source=/dev/null
```
edmorley: I think this can be `source=bin/steps/hooks/pre_compile` ? (And same for the below)

Makefile
```diff
@@ -3,6 +3,11 @@
 
 test: test-heroku-16
 
+check:
+	@shellcheck bin/compile bin/detect bin/release bin/test-compile bin/utils bin/warnings
```
edmorley: One option would be to use git-grep to search for relevant shebangs to avoid having to keep this list up to date, like so:
https://github.com/mozilla-platform-ops/devservices-aws/blob/1ea2f0b28d5ef23885b451873129e084d287be2b/runtests.sh#L17

bin/compile
```diff
@@ -55,10 +59,12 @@ unset RECEIVE_DATA RUN_KEY BUILD_INFO DEPLOY LOG_TOKEN
 unset CYTOKINE_LOG_FILE GEM_PATH
 
 # Syntax sugar.
-source $BIN_DIR/utils
+# shellcheck source=/dev/null
```
dzuelke: Why this, and not the correct path? That's what the comment is for, and you want the import to occur, so that Shellcheck can follow it

bin/compile
dzuelke: Yea

bin/compile
```diff
@@ -92,148 +98,170 @@ if [[ ! -f Procfile ]]; then
...
 
 
 # Django collectstatic support.
 let start=$(nowms)
-sub-env $BIN_DIR/steps/collectstatic
+sub-env "$BIN_DIR/steps/collectstatic"
 mtime "collectstatic.time" "${start}"
 
 # Create .profile script for application runtime environment variables.
+# shellcheck disable=SC2016
```
dzuelke: Maybe backslash escapes instead of single quotes; it might not complain then?

bin/steps/nltk
```diff
@@ -10,18 +10,23 @@
...
 # Syntax sugar.
+# shellcheck source=/dev/null
 source $BIN_DIR/utils
 
 # Check that nltk was installed by pip, otherwise obviously not needed
-python -m nltk.downloader -h >/dev/null 2>&1
-if [ $? -eq 0 ]; then
+if ! python -m nltk.downloader -h >/dev/null 2>&1; then
     puts-step "Downloading NLTK corpora..."
+
```
dzuelke: Are all those blank lines intentional?

bin/steps/pipenv-python-version
```diff
@@ -8,14 +10,15 @@ if [[ -f $BUILD_DIR/Pipfile.lock ]]; then
         fi
         if [[ -f $BUILD_DIR/Pipfile.lock ]]; then
             set +e
-            PYTHON=$(cat $BUILD_DIR/Pipfile.lock | jq '._meta.requires.python_version' -r)
+            # shellcheck disable=SC2002
```
dzuelke: Just do `PYTHON=$(jq -r '._meta.requires.python_version' "$BUILD_DIR/Pipfile.lock")` instead!?

bin/steps/python
```diff
@@ -42,7 +42,7 @@ if [ ! "$SKIP_INSTALL" ]; then
 fi
 
 # If Pip isn't up to date:
-if [ "$FRESH_PYTHON" ] || [[ ! $(pip --version) == *$PIP_VERSION* ]]; then
+if [ "$FRESH_PYTHON" ] || [[ ! $(pip --version) == *$PIP_UPDATE* ]]; then
```
dzuelke: I fail to see what this change fixes. You simply renamed a variable?!

bin/compile
kennethreitz: I couldn't get that to work, for some reason. I'll re-attempt, however.

bin/steps/nltk
kennethreitz: making it more readable

bin/compile
kennethreitz: i tried that, but it was still erroring out for me. I'll try again.

kennethreitz:
@dzuelke all commits will be squashed on merge

bin/compile
kennethreitz: ah, the cli is fine, but my sublime plugin is unhappy. interesting.

dzuelke: Probably a base/working dir problem in Sublime, since the path is relative?

bin/steps/pipenv-python-version
kennethreitz: ✨🍰✨

bin/compile
kennethreitz: Success!

Makefile
kennethreitz: I don't mind, the files don't change often at all.

bin/compile
kennethreitz: I think it's because sublime isn't running shellcheck with the `-x` flag passed.

bin/compile
kennethreitz: ✨🍰✨

kennethreitz:
k, ready for re-review.

kennethreitz:
One test is failing — I'm investigating, but don't let that stop approval.

kennethreitz:
Tests should be passing now! ✨🍰✨

bin/steps/python
kennethreitz: this fixes a bug with pip.

kennethreitz: turns out, pip completely fails to run if PIP_VERSION is set and isn't equal to what is installed.

bin/compile
```diff
@@ -92,149 +98,168 @@ if [[ ! -f Procfile ]]; then
...
 # Install sane-default script for $WEB_CONCURRENCY and $FORWARDED_ALLOW_IPS.
-cp $ROOT_DIR/vendor/WEB_CONCURRENCY.sh $WEB_CONCURRENCY_PROFILE_PATH
-cp $ROOT_DIR/vendor/python.gunicorn.sh $GUNICORN_PROFILE_PATH
+cp "$ROOT_DIR/vendor/WEB_CONCURRENCY.sh" "$WEB_CONCURRENCY_PROFILE_PATH"
+cp "$ROOT_DIR/vendor/python.gunicorn.sh" "$GUNICORN_PROFILE_PATH"
 
 
 # Experimental post_compile hook.
-source $BIN_DIR/steps/hooks/post_compile
+# shellcheck source=bin/steps/hooks/post_compile
```
hunterloftis: what are these source comments?

edmorley:
> turns out, pip completely fails to run if PIP_VERSION is set and isn't equal to what is installed.

Yeah I guess for the same reason as:
https://github.com/docker-library/python/blob/dc08012bd82a2c566abd1c1150554845cb749281/2.7/jessie/Dockerfile#L57-L58

bin/steps/python
```diff
@@ -24,37 +26,35 @@ if [ ! "$SKIP_INSTALL" ]; then
...
 fi
 
 # If Pip isn't up to date:
-if [ "$FRESH_PYTHON" ] || [[ ! $(pip --version) == *$PIP_VERSION* ]]; then
-  WORKING_DIR=$(pwd)
+if [ "$FRESH_PYTHON" ] || [[ ! $(pip --version) == *$PIP_UPDATE* ]]; then
 
   TMPTARDIR=$(mktemp -d)
-  trap "rm -rf $TMPTARDIR" RETURN
+  trap 'rm -rf $TMPTARDIR' RETURN
```
edmorley: I don't think the `TMPTARDIR` directory is actually used any more? (Was previously used for the manual pip installation before the switch to get-pip.py)

kennethreitz: ah, nice catch!

bin/compile
kennethreitz: they tell shellcheck how to check sources that come from variables. standard practice.

PR #442, opened 2017-08-04
Improve UX when clearing cache due to the stack changing

edmorley:
**1) Adds a test to check that the cache is invalidated when the stack changes.**

**2) Improves the UX when clearing cache due to the stack changing.**
Now outputs a message informing that the cache was cleared, and clears the cache first to avoid a redundant message about removing an old Python version.


kennethreitz:
this looks solid!

PR #443, opened 2017-08-04
Speed up the multi-stage Travis run

edmorley:
Reduces the impact on Travis runtime caused by #438, by:
* Making the shellcheck job run on the container infra (`sudo: false`) which has faster boot times.
* Running the heroku-16 and cedar-14 stack tests in parallel, since previously they were separate stages (which are run in series by design).

Reduces Travis end-to-end time from ~10 minutes to ~6 minutes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,23 @@
 language: bash
-# sudo: required
-addons:
-  apt:
-    sources:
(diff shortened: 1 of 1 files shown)


kennethreitz:
✨🍰✨

kennethreitz:
will merge once tests pass

PR #449, opened 2017-08-16
Adding libjasper.so to fix libgdal dependency issue on heroku-16 stacks

zachmccormick:
libjasper is present during the build phase, but is not ever actually added to the slug, causing loading/linking to fail whenever you try to load libgdal (rgeo or GeoDjango). This addition copies the file to the vendor libs directory and solves the problem!


zachmccormick:
Additionally, the pages at https://devcenter.heroku.com/articles/python-c-deps and https://devcenter.heroku.com/articles/postgis should be updated to reflect the need (at least for the latest stable Django) to add the following to `settings.py`:

```python
GDAL_LIBRARY_PATH = os.getenv('GDAL_LIBRARY_PATH')
GEOS_LIBRARY_PATH = os.getenv('GEOS_LIBRARY_PATH')
```

and set the environment variables to
```
GDAL_LIBRARY_PATH=/app/.heroku/vendor/lib/libgdal.so
GEOS_LIBRARY_PATH=/app/.heroku/vendor/lib/libgeos_c.so
```

kennethreitz:
I believe `GDAL_LIBRARY_PATH` and `GEOS_LIBRARY_PATH` get set but init script, but I'll have to confirm.

kennethreitz:
✨🍰✨

zachmccormick:
Thanks! Feel free to ping me if there are any other weird C/C++ library issues related to the Python buildpack. I'm always happy to help debug those otherwise terrible issues - I always feel like I learn new things about binaries, compilers, and platform nuances!

kennethreitz:
Updated the docs, pushing a new release of the buildpack.

kennethreitz:
@zachmccormick what's a good way to test that this is working now? e.g. a django function to call.

PR #460, opened 2017-09-05
NLTK support: Fix passing of multiple corpora identifiers

### LINKED ISSUE #444: Error downloading NLTK corpora
Hi, we are no longer able to download and install the NLTK corpora via this build pack.

We receive the deployment error below, no matter the packages listed in the `nltk.txt` file:

```
remote: -----> Downloading NLTK corpora...
remote: -----> Downloading NLTK packages: wordnet brown punkt averaged_perceptron_tagger 
remote:        [nltk_data] Error loading wordnet brown punkt
remote:        [nltk_data]     averaged_perceptron_tagger : Package 'wordnet brown
remote: Traceback (most recent call last):
remote:   File "/app/.heroku/python/lib/python2.7/runpy.py", line 174, in _run_module_as_main
remote:     "__main__", fname, loader, pkg_name)
remote:   File "/app/.heroku/python/lib/python2.7/runpy.py", line 72, in _run_code
remote:     exec code in run_globals
remote:        [nltk_data]     punkt averaged_perceptron_tagger ' not found in index
remote:        Error installing package. Retry? [n/y/e]
remote:   File "/tmp/build_5f9f482255a2eae0e0f03f81ed130b83/.heroku/python/lib/python2.7/site-packages/nltk/downloader.py", line 2267, in <module>
remote:     halt_on_error=options.halt_on_error)
remote:   File "/tmp/build_5f9f482255a2eae0e0f03f81ed130b83/.heroku/python/lib/python2.7/site-packages/nltk/downloader.py", line 675, in download
remote:     choice = compat.raw_input().strip()
remote: EOFError: EOF when reading a line
```

edmorley:
**1) Update NLTK test to use multiple corpora**
So that the incorrect handling of multiple IDs seen in #444 would have been caught. Also switches to some of the smaller corpora, to reduce time spent downloading during tests (see sizes on http://www.nltk.org/nltk_data/).

Example failing CI job with the updated test but without the fix below:
https://travis-ci.org/edmorley/heroku-buildpack-python/builds/271986505

**2) Fix passing of multiple NLTK corpora identifiers**
As part of fixing the shellcheck warnings in #438, double quotes had been placed around `$nltk_packages` passed to the `nltk.downloader`, which causes multiple identifiers to be treated as though it were just one identifier that contains spaces - which results in the downloader failing to find them.

The docs for the shellcheck warning in question recommend using arrays if the intended behaviour really is to split on spaces:
https://github.com/koalaman/shellcheck/wiki/SC2086#exceptions

As such, `readarray` has been used, which is present in bash >=4.

The `[*]` array form is used in the log message, to prevent shellcheck warning [SC2145](https://github.com/koalaman/shellcheck/wiki/SC2145), whereas `[@]` is used when passed to `nltk.downloader` to ensure the array elements are unpacked as required.

Note: Both before and after this fix, using anything but unix line endings in `nltk.txt` will also cause breakage. I'm happy to add a fix for that too if Windows CRLF line endings are something that should be supported.

Fixes #444.

### DIFF
--- a/bin/steps/nltk
+++ b/bin/steps/nltk
@@ -21,10 +21,10 @@ if sp-grep -s nltk; then
 
     if [ -f "$nltk_packages_definition" ]; then
 
-        nltk_packages=$(tr "\n" " " < "$nltk_packages_definition")
-        puts-step "Downloading NLTK packages: $nltk_packages"
+        readarray -t nltk_packages < "$nltk_packages_definition"
+        puts-step "Downloading NLTK packages: ${nltk_packages[*]}"
 
-        python -m nltk.downloader -d "$BUILD_DIR/.heroku/python/nltk_data" "$nltk_packages"  | indent
+        python -m nltk.downloader -d "$BUILD_DIR/.heroku/python/nltk_data" "${nltk_packages[@]}" | indent
         set_env NLTK_DATA "/app/.heroku/python/nltk_data"
 
     else
@@ -32,4 +32,3 @@ if sp-grep -s nltk; then
         puts-warn "Learn more: https://devcenter.heroku.com/articles/python-nltk"
     fi
 fi
-
--- a/test/fixtures/nltk/nltk.txt
+++ b/test/fixtures/nltk/nltk.txt
@@ -1,1 +1,2 @@
-wordnet
+city_database
+stopwords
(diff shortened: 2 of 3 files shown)


kennethreitz:
\o/

edmorley:
I realised I forgot to add a changelog entry, I don't suppose you could add one as part of #466?
(I'd do it myself, but it will only conflict with the changelog changes there).

kennethreitz:
yessir!

Issue #448, opened 2017-08-15
Git tag missing for the v110 release


edmorley:
https://github.com/heroku/heroku-buildpack-python/blob/master/CHANGELOG.md#110
https://github.com/heroku/heroku-buildpack-python/releases

edmorley:
The v111 and v112 tags are missing too.

The lack of these tags makes it less straight-forwards for people to roll back to a previous buildpack release in case of issues with the `heroku/python` stable release (eg #457).

kennethreitz:
This is a side-effect of our new master branch protection, which gave me a mental block from pushing to the master branch (including pushing git tags — even though that's allowed). Sorry for the trouble.

kennethreitz:
Just pushed a tag for v112.

kennethreitz:
TODO:

- [ ] Add tag for v110
- [ ] Add tag for v111

kennethreitz:
Need to find the right refs first.

edmorley:
The new v115 release's tag is missing too (it should be on master, which is at d4b8c2aeda335c48b9f7dd6731c4b406d319ae1e).
I'll file an issue for adding a way to automate this :-)

> Need to find the right refs first.

By modifying the buildkits plugin so it dumps the raw JSON response from `heroku buildkits:revisions heroku/python`, I found an additional `tar_link` property not normally output, which gives the previous buildpack URLs as:
https://codon-buildpacks.s3.amazonaws.com/buildpacks/heroku/python-v110.tgz
https://codon-buildpacks.s3.amazonaws.com/buildpacks/heroku/python-v111.tgz

By diffing those against each other and comparing to the repository git log, I believe v111 should be at fb52d1bf595a1e4735d54724ffb778793ab64d38 , and v110 at 194932889dee79c1a6045e6208cca40edb90147c.

It also means that the v109/v110 CHANGELOG.md entries aren't quite right - I'll open a PR for that now.

kennethreitz:
@edmorley you're a gentleman and a scholar. 

Pushed tags.

edmorley:
Looks great - many thanks :-)

PR #469, opened 2017-09-06
Correct changelog entries for v108, v109 and v110

edmorley:
Found by diffing the published buildpack archives and looking at the git log to see what changed, as part of trying to work out the correct git refs for #448.

It looks like the v108 entry was accidentally added in #405, when the release being published there was actually v107. The setuptools update mentioned never happened, since the `SETUPTOOLS_VERSION` variable was found to be unused and instead removed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,17 +28,17 @@ Linting, bugfixes.
 
 # 110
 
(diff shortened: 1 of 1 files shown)


kennethreitz:
✨🍰✨

PR #473, opened 2017-09-22
Allow pipenv version override

jacebrowning:
My Django app started failing to deploy today:

```
-----> Python app detected
-----> Running pre-compile hook
-----> Installing python-3.6.1
-----> Installing pip
-----> Installing requirements with latest pipenv...
       No package provided, installing all dependencies.
       Installing dependencies from Pipfile.lock…
-----> Running post-compile hook
-----> Collecting static files...
       Traceback (most recent call last):
         File "manage.py", line 6, in <module>
           from django.core.management import execute_from_command_line
       ModuleNotFoundError: No module named 'django'
 !     Push failed
```

At least while pipenv is under rapid development, I think there should be an option to disable the `--upgrade` option when pipenv is installed during builds.

### DIFF
--- a/bin/steps/pipenv
+++ b/bin/steps/pipenv
@@ -3,10 +3,15 @@
 # Pipenv support (Generate requriements.txt with pipenv).
 if [[ -f Pipfile ]]; then
     if [[ ! -f requirements.txt ]]; then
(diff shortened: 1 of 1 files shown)


kennethreitz:
we'll start pinning it

PR #478, opened 2017-10-01
Fix #398 -- Set explict GDAL and GEOS location

### LINKED ISSUE #398: Django does find GEOS libraries on Heroku-16
I'm getting an `ImportError` from Django when switching to the heroku-16 stack.

```
File "/app/.heroku/python/lib/python3.6/site-packages/django/contrib/gis/geos/libgeos.py", line 57, in load_geos
'", "'.join(lib_names)
ImportError: Could not find the GEOS library (tried "geos_c", "GEOS"). Try setting GEOS_LIBRARY_PATH in your settings.
```

After some digging, I don't see the difference. The libraries are installed and the `LD_LIBRARY_PATH` is still the same as it used to be on cedar-14.

I could fix the issue for me by setting `GEOS_LIBRARY_PATH` explicitly in the `post_compile` hook.

`bin/post_compile`
```shell
#!/usr/bin/env bash

set -e

export GEOS_LIBRARY_PATH="$BUILD_DIR/.heroku/vendor/lib"
```

codingjoe:
Set GDAL and GEOS library locaiton explicitly in environment
variables:

*   `GDAL_LIBRARY_PATH="/app/.heroku/vendor/lib/libgdal.so"`
*   `GEOS_LIBRARY_PATH="/app/.heroku/vendor/lib/libgeos_c.so"`

Django has to settings with the same name. The setup now works as
described here:
https://devcenter.heroku.com/articles/postgis#geodjango-setup

### DIFF
--- a/bin/steps/geo-libs
+++ b/bin/steps/geo-libs
@@ -40,5 +40,10 @@ if [[ "$BUILD_WITH_GEO_LIBRARIES" ]]; then
 
   GDAL=$(pwd)/vendor
   export GDAL
+  # set path for post_compile hooks
+  export GDAL_LIBRARY_PATH="$BUILD_DIR/.heroku/vendor/lib/libgdal.so"
+  export GEOS_LIBRARY_PATH="$BUILD_DIR/.heroku/vendor/lib/libgeos_c.so"
+  # set path for runtime environmeht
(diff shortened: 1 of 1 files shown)


codingjoe:
@kennethreitz is this happening any time soon? As is GeoDjango das not work on heroku-16 without this patch :/

kennethreitz:
✨🍰✨

PR #481, opened 2017-10-06
Add PyPy 5.9.0

joshfriend:
Adds support for Pandas/NumPy! https://morepypy.blogspot.com/2017/10/pypy-v59-released-now-supports-pandas.html


tdbear:
So waiting for it...

kennethreitz:
Pypy has always been documented as "experimental". While merging this pull request seems easy, there's a lot of manual steps involved in deploying this to production — and with the extremely low usage that pypy gets on the platform, it's simply not worth the effort. 

I'm open to pushing new versions occasionally, but this is the first instance where we're documenting that we don't plan to support every version of Pypy.

PR #479, opened 2017-10-03
Add CPython 3.6.3

joshfriend:

PR #483, opened 2017-10-26
update latest python 3 to 3.6.3

AlJohri:


AlJohri:
seems like #479 is more correct

Issue #453, opened 2017-08-22
Output the Python version when restoring from cache


edmorley:
In #440 it was mentioned that 99% of apps don't specify a Python version via a checked-in runtime.txt and instead rely on the default Python version provided by the buildpack. 

In addition, the buildpack uses a sticky version -- that is whilst the buildpack default Python version does increase over time, an app will always use the version that was the default when it was first created [1].

This means apps can be using ancient/insecure [2] versions of Python with users having no way to notice, since the buildpack log output doesn't mention what version of Python is being used when it is restored from cache (unlike first install). 

Log output when Python restored from cache:
```
-----> Python app detected
-----> Installing requirements with pip
```

The buildpack API docs [recommend](https://devcenter.heroku.com/articles/buildpack-api#style) including version numbers where possible, and other buildpacks such as the nodejs one provide much more detailed information. 

nodejs buildpack output:
```
-----> Installing binaries
       engines.node (package.json):  7.10.0
       engines.npm (package.json):   unspecified (use default)
       engines.yarn (package.json):  unspecified (use default)
       
       Resolving node version 7.10.0...
       Downloading and installing node 7.10.0...
       Using default npm version: 4.2.0
       Resolving yarn version (latest)...
       Downloading and installing yarn (0.28.1)...
       Installed yarn 0.28.1
```

[1] If we're being pedantic, it in fact uses the default version when the apps' cache was last cleared, which may be more recent than when the app was created. But I doubt this set of people would ever manually clear their cache either, so it's the same end result.
[2] Yes, there is the warning for `InsecurePlatformWarning`, but that only covers Python <2.7.9 and there have been other security fixes since.

kennethreitz:
Closing for #440

PR #484, opened 2017-11-07
Output the found Python version when restoring from cache

edmorley:
This makes it clearer which Python version is in use (for people who are not pinning via `runtime.txt`), plus that a cache is being used (for people debugging and wondering why their eg review app has different behaviour from an existing app).

Also:
* makes the conditional clearer by inverting to avoid the negation
* avoids the duplicated `cat .heroku/python-version`
* uses bash quoting and test `[[` best practices
* adds a test for the Python version changing (previously untested)

Refs #453.

### DIFF
--- a/bin/steps/python
+++ b/bin/steps/python
@@ -25,11 +25,13 @@ if [[ "$STACK" != "$CACHED_PYTHON_STACK" ]]; then
 fi
 
 if [ -f .heroku/python-version ]; then
-  if [ ! "$(cat .heroku/python-version)" = "$PYTHON_VERSION" ]; then
-      puts-step "Found $(cat .heroku/python-version), removing"
-      rm -fr .heroku/python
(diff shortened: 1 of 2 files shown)


kennethreitz:
I'd like to avoid this verbosity.

edmorley:
@kennethreitz, it's one extra line of output - compared to the 6 or so that the nodejs buildpack uses to display versions - and I believe would help users debug issues when comparing environments (eg a new stage environment or review app that ends up using a different version of Python).

jmorrell:
I think this deserves more discussion.

I'd love it if the build push experience were more designed and less verbose, but given what we currently have, we should tend to being explicit about what's happening.

kennethreitz:
Our current workflow is quite good, in my opinion.

kennethreitz:
Will def consider re-opening in the future.

PR #480, opened 2017-10-03
Default to python 3.6.3

joshfriend:
Made this separate from #479 so you can reject it separately if I did something wrong, or you don't want to update the default yet :)

TODO:
* [x] Merge #479

PR #485, opened 2017-11-14
Update default to latest (Python 3.6.3)

kennethreitz:


edmorley:
This is a dupe of #480 (and is missing a few changes from there)

PR #487, opened 2017-11-26
Added support for compiling .coco [coconut-lang] files

creatorrr:
Added support for compiling `.coco` files in order to support [coconut-lang](https://coconut.readthedocs.io/en/master/DOCS.html). The compilation step is run in `post_compile` hook and the compiled files are cached.


kennethreitz:
this would be better as a separate buildpack

PR #490, opened 2017-12-11
Make gunicorn default access log to stdout

kennethreitz:
This pull request allows gunicorn to log to stdout for all incoming requests, which it does not by default.

This was not previously an option.


edmorley:
Good idea! I'd thought about manually enabling this for some of our gunicorn based apps previously - since whilst it means doubling up on the Heroku router log output, it includes additional metadata (such as User Agent, [see here](http://docs.gunicorn.org/en/stable/settings.html#access-log-format)) that isn't present in the former.

PR #494, opened 2017-12-26
Add pypy[23] 5.10

joshfriend:
Last time when I submitted PyPy 5.9, you declined to merge it because it was ["experimental"](https://github.com/heroku/heroku-buildpack-python/pull/481#issuecomment-349975244), which was fair (at least for the 3.5 version, which was considered "beta")

PyPy3.5 is [no longer in beta](https://morepypy.blogspot.com/2017/12/pypy27-and-pypy35-v510-dual-release.html)

### DIFF
--- a/builds/runtimes/pypy-5.10
+++ b/builds/runtimes/pypy-5.10
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+# Build Path: /app/.heroku/python/
+# Build Deps: libraries/sqlite
+
+OUT_PREFIX=$1
+
+echo "Building PyPy…"
+SOURCE_TARBALL='https://bitbucket.org/pypy/pypy/downloads/pypy2-v5.10.0-linux64.tar.bz2'
+curl -L $SOURCE_TARBALL | tar jx
+cp -R pypy2-v5.10.0-linux64/* $OUT_PREFIX
+
+ln $OUT_PREFIX/bin/pypy $OUT_PREFIX/bin/python
--- a/builds/runtimes/pypy3-5.10
+++ b/builds/runtimes/pypy3-5.10
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+# Build Path: /app/.heroku/python/
+# Build Deps: libraries/sqlite
+
+OUT_PREFIX=$1
+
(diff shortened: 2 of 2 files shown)


kennethreitz:
still no plans to update pypy at this time, but maybe next time around ;)

joshfriend:
Sounds good, see you next release ;)

edmorley:
I don't personally use PyPy, but I was thinking that perhaps another solution here might be to switch it to compiling at app build time, then caching the result of the build the same as the extracted archive would be? The very first build for a new app using PyPy (or after each version bump) would be slow, however there would be no difference thereafter. How long does the PyPy compile take?

joshfriend:
I've compiled pypy from source a few times and it took **well over an hour** on my macbook pro with a 2.0GHz quad-core i7. PyPy provides suitable binaries anyways so I think it's rather pointless to compile from source (in the _vast_ majority of cases).

edmorley:
Oh sorry I'd missed that unlike for CPython, the existing PyPy scripts in this repo only have to extract the archive rather than compile anything. In which case the idea of doing everything at app build time is even more viable. 

ie: I propose adjusting `bin/steps/python` such that for PyPy only, the steps typically performed by `builds/runtimes/pypy-*` are instead run at app build time, rather than being an out-of-band step performed by Kenneth.

This would mean:
* no code changes required (or archives uploaded to S3) to support new PyPy versions, so no more trying to convince Kenneth to support them each time
* minimal impact on app compile time for the clean-cache case, since there's not even a compile to worry about. (The only difference being speed/uptime differences between `https://bitbucket.org/pypy/pypy/downloads/...` and `https://lang-python.s3.amazonaws.com/...`)
* zero impact on app compile time for the 2nd build onwards, since the build cache will be used as before

kennethreitz:
@edmorley i like this idea!

joshfriend:
since the official buildpack supports pipenv now, is there any reason it can't also use pipenv's integration with pyenv to install the requested python version?

kennethreitz:
Yes, that is an expensive operation.

kennethreitz:
Expensive meaning time-consuming. Also, our builds are better than the ones pyenv provides.

PR #628, opened 2018-01-10
Update python.gunicorn.sh

kennethreitz:
This allows customers to override this environment variable.


edmorley:
(Comment updated above to correct testing steps - looks like setting to an empty string can't override the access log setting after all - perhaps this is fine)

PR #637, opened 2018-02-03
Fix PYTHONPATH is not set when running collectstatic

jxltom:
In command ```bin/steps/collectstatic```, PYTHONPATH is not set.

If manage.py file is not located at root folder but in the subfolder of a project and if ```import some_module_in_project``` is called in ```manage.py```, ```python somefolder/manage.py collectstatic --noinput``` will raise ```can not find module``` error. This is due to ```PYTHONPATH``` is not set for root folder and python can not import project's own module.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Python Buildpack Changelog
 
+# 125
+
+Set `PYTHONHOME` during collectstatic runs.
+
 # 124
 
 Update buildpack to automatically install [dev-packages] (Pipenv) during Heroku CI builds.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 # 125
 
-Set `PYTHONHOME` during collectstatic runs.
+Set `PYTHONPATH` during collectstatic runs.
 
 # 124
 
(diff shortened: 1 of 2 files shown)


jxltom:
I found this issue because I have a customized ```manage.py``` which will do some additional work for me. But this ```manage.py``` needs import some modules of the project and since PYTHONPATH is not set, ```python manage.py collectstatic``` will raise can not find module error

Here is the project layout

```
Procfile
requirements.txt
myapplication
    manage.py
    settings.py
    ....other...stuff...
```

kennethreitz:
thanks!

jxltom:
Glad to contribute!

jxltom:
@kennethreitz It looks like buildpack in Heroku is cached. May I know how can I update the buildpack in Heroku? It looks like the buildpack is not using the latest one in github even I create a new project in Heroku

edmorley:
The `heroku/python` buildpack alias maps to stable releases rather than master of this repo. The versions/release dates can be seen using:
https://github.com/heroku/heroku-buildkits#revisions

In the meantime if you wish to use master, set the buildpack URL to this git repository like so:
https://devcenter.heroku.com/articles/buildpacks#using-a-third-party-buildpack

PR #638, opened 2018-02-03
Account for the root not being /app

Photonios:
From what I understood, the buildpack creates symlinks in `/app`, during build so that the path to Python shorter and the shebang in for example `pip` do not exceed the maximum length of 127 characters. This is smart little work-around, but this causes a problem with buildpacks that make other buildpacks run in a sub-directory of `/app`. The symlinks are set up during the build, but are lost when the slug gets deployed. This small modification does two things:

* Set up symlinks at runtime as well if the `$HOME` is not `/app`.
* Always use `$HOME` in `export` statements in the `.profile.d` script.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -224,14 +224,24 @@ mtime "collectstatic.time" "${start}"
 # Create .profile script for application runtime environment variables.
 set_env PATH "\$HOME/.heroku/python/bin:\$PATH"
 set_env PYTHONUNBUFFERED true
-set_env PYTHONHOME /app/.heroku/python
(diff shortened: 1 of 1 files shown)


kennethreitz:
I like the idea of using `$HOME` instead of `/app`. I'm not sold about the idea of making `/app` a symlink during runtime, however.

Photonios:
@kennethreitz Do you have alternative for the symlinks? If you set them up during the build, then how would paths still work at runtime?

kennethreitz:
this seems like a correct approach, for other buildpack environments.

PR #643, opened 2018-02-16
feat: allow stack url overrides using BUILDPACK_VENDOR_URL env var

josegonzalez:
This is useful if you'd like to customize python binaries without forking the entire buildpack.

Similar functionality was implemented in https://github.com/heroku/heroku-buildpack-ruby/pull/238

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -33,6 +33,12 @@ ENV_DIR=$3
 
 export BUILD_DIR CACHE_DIR ENV_DIR
 
+VENDOR_URL="https://lang-python.s3.amazonaws.com/$STACK"
+if [[ -n ${BUILDPACK_VENDOR_URL:-} ]]; then
+    VENDOR_URL="$BUILDPACK_VENDOR_URL"
+fi
+export VENDOR_URL
+
 # Python defaults
 DEFAULT_PYTHON_VERSION="python-3.6.4"
 LATEST_3="python-3.6.4"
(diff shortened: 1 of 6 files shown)


jxltom:
I am curious that will it be better when ```BUILDPACK_VENDOR_URL``` does not includes ```STACK``` part.

jxltom:
Actually ```https://lang-common.s3.amazonaws.com/buildpack-stdlib/v8/stdlib.sh``` is also hardcoded https://github.com/heroku/heroku-buildpack-python/blob/9468ec263012daed0eb16e689ad7d7cf892a1e22/bin/utils#L7. How about provide a URL such as BUILDPACK_BASE_URL for replacing all hardcoded amazon s3 url at once?

kennethreitz:
thanks for this! unfortunately, this won't on heroku as-is. environment variables aren't exported to the buildpack automatically, and need to be loaded with some utility functions. 

is this for use off of heroku?

josegonzalez:
Sort of. I'm working with a client that has a few of their applications on Heroku but does their CI on GCP (Yes, I'm aware of the Heroku CI offering). I'd like to provide a way to add CI support against the underlying buildpacks, but every so often see network connectivity issues from GCP to AWS, so being able to vendor those on GCP's blob store would be nice.

josegonzalez:
Thanks @kennethreitz !

PR #656, opened 2018-03-13
don't skip Pipenv installation if there are git deps

### LINKED ISSUE #654: Generated `requirements.txt` doesn't include git repo deps
Here is a part of my `Pipfile`
```
[packages]
...
gunicorn = "==19.7.1"
flask-login = "==0.4.0"
"psycopg2" = "*"
mistune-contrib = {git = "https://github.com/frostming/mistune-contrib.git", editable = true}
flask-sslify = "*"
```
After deployment, the app can't be established because it can't find `mistune-contrib`. I login `heroku run bash` to see the generated `requirements.txt` and found that mistune-contrib is not included

This issue blocks my deployment.

kennethreitz:
Fixes #654

Signed-off-by: Kenneth Reitz <me@kennethreitz.org>

### DIFF
--- a/bin/steps/pipenv
+++ b/bin/steps/pipenv
@@ -11,7 +11,7 @@ if [[ -f Pipfile.lock ]]; then
         if [[ $(openssl dgst -sha256 Pipfile.lock) == $(cat .heroku/python/Pipfile.lock.sha256) ]]; then
 
             # Don't skip installation of there are git deps.
-            if grep -q 'git' Pipfile.lock; then
+            if ! grep -q 'git' Pipfile.lock; then
                 echo "Skipping installation, as Pipfile.lock hasn't changed since last deploy." | indent
                 export SKIP_PIPENV_INSTALL=1
                 export SKIP_PIP_INSTALL=1
--- a/bin/steps/pipenv
+++ b/bin/steps/pipenv
@@ -9,6 +9,8 @@ set -e
 if [[ -f Pipfile.lock ]]; then
     if [[ -f .heroku/python/Pipfile.lock.sha256 ]]; then
         if [[ $(openssl dgst -sha256 Pipfile.lock) == $(cat .heroku/python/Pipfile.lock.sha256) ]]; then
+            # Measure that we're using Pipenv.
+            mcount "tool.pipenv"
 
             # Don't skip installation of there are git deps.
             if ! grep -q 'git' Pipfile.lock; then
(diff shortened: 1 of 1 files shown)


bin/steps/pipenv
```diff
@@ -9,14 +9,14 @@ set -e
 if [[ -f Pipfile.lock ]]; then
     if [[ -f .heroku/python/Pipfile.lock.sha256 ]]; then
         if [[ $(openssl dgst -sha256 Pipfile.lock) == $(cat .heroku/python/Pipfile.lock.sha256) ]]; then
-            if [[ ! "$PIPENV_ALWAYS_INSTALL" ]]; then
+
+            # Don't skip installation of there are git deps.
+            if ! grep -q 'git' Pipfile.lock; then
```
mjtamlyn: Do you want to grep for `git = ` instead of just `git`? This makes it a bit more likely it will just reference editable checkouts, rather than any of the 891 packages on PyPI which include the string `git` in their name (e.g. `django-logit`, `django-taggit`, anything referencing `digitalocean`...)

kennethreitz: yeah, i wasn't too worried about it, as the only opt-in here is proceeding to install

kennethreitz: problem is python might use `git = `, `git=`, `"git"  =  `

kennethreitz: I'd rather catch *all* genuine git deployments, than miss some, based on Pipfile syntax.

bin/steps/pipenv
```diff
@@ -9,14 +9,16 @@ set -e
 if [[ -f Pipfile.lock ]]; then
     if [[ -f .heroku/python/Pipfile.lock.sha256 ]]; then
         if [[ $(openssl dgst -sha256 Pipfile.lock) == $(cat .heroku/python/Pipfile.lock.sha256) ]]; then
-            if [[ ! "$PIPENV_ALWAYS_INSTALL" ]]; then
+            # Measure that we're using Pipenv.
+            mcount "tool.pipenv"
+
+            # Don't skip installation of there are git deps.
+            if ! grep -q 'git' Pipfile.lock; then
```
jmorrell: This would trigger even if a module name just contained `git` no?

jmorrell: Also tests seem like they would be useful here

kennethreitz: this is a hotfix, needs to be released ASAP. Current testing framework doesn't allow for re-builds (which would trigger this). See above comments for justification on using 'git'

edmorley: > Current testing framework doesn't allow for re-builds (which would trigger this)

It's possible by passing in the cache directory manually, like these tests do:
https://github.com/heroku/heroku-buildpack-python/blob/7d975e74a93077f301916c85bce1089d59cb11f6/test/run#L85-L104

kennethreitz: happy to do that after releasing this — this is breaking customer builds at the moment.

kennethreitz: e.g. *emergency*

PR #657, opened 2018-03-13
Metrics

kennethreitz:


bin/steps/collectstatic
```diff
@@ -39,6 +39,8 @@ if [ ! "$DISABLE_COLLECTSTATIC" ] && [ -f "$MANAGE_FILE" ] && [ "$DJANGO_INSTALL
     # Display a warning if collectstatic failed.
     [ "$COLLECTSTATIC_STATUS" -ne 0 ] && {
 
+        mcount "failure.collectstatic"
```
jmorrell: Can you break this down any more? What are some common ways that collect static fails?

Ex: 
```
django.core.exceptions.ImproperlyConfigured
SyntaxError: invalid syntax
OSError: [Errno 2] No such file or directory
SyntaxError: Missing parentheses in call to 'print'
The CSS file 'css/global.css' references a file which could not be found
```

kennethreitz: it fails in hundreds of ways. I'd like to start measuring it period, then work from there.

PR #660, opened 2018-03-14
collectstatic metrics

kennethreitz:
This pull request adds additional metrics to collectstatic failures.


bin/steps/collectstatic
```diff
@@ -42,6 +46,22 @@ if [ ! "$DISABLE_COLLECTSTATIC" ] && [ -f "$MANAGE_FILE" ] && [ "$DJANGO_INSTALL
 
         mcount "failure.collectstatic"
 
+        if grep -q 'SyntaxError' "$COLLECTSTATIC_LOG"; then
+            mcount "failure.collectstatic.syntax-error"
```
jmorrell: You might want to turn these into `else if`s if they are mutually exclusive

kennethreitz: they aren't.

kennethreitz: python can have multiple tracebacks

PR #662, opened 2018-03-16
Output pipenv version in steps

zyv:
Feels very useful to me, making it much easier to track which version of pipenv exactly is being used, since it's being pinned now. Hope you'll also see it this way!


kennethreitz:
✨🍰✨

PR #671, opened 2018-03-23
Move SKIP_PIP_INSTALL=1 after pip-uninstall step

jxltom:
Currently, deployment via pipenv won't uninstall requirements-stale.txt...

```export SKIP_PIP_INSTALL=1``` should be after ```"$BIN_DIR/steps/pip-uninstall"``` otherwise pip-uinstall step will be bypassed.


kennethreitz:
Will evaluate out of band. This part of the codebase is very tricky, and I'm very apprehensive to accept any external changes to it.

PR #677, opened 2018-04-02
check for locale before setting in test-compile

zeroshade:
When on systems without C.UTF-8 locale installed, the fact that LC_ALL is set to it causes CI to not work correctly so we should check for it existing first before setting LC_ALL. Specifically, if any of the python dependencies pull in https://pypi.python.org/pypi/click then installation fails due to requiring proper unicode support.


kennethreitz:
I think that locale is simply a requirement of using this buildpack, as it is a requirement of using pipenv.

Issue #678, opened 2018-04-04
pipenv-to-pip fails for python3 older than 3.6


pdamodara:
https://github.com/heroku/heroku-buildpack-python/blob/9eddeaeef98dbbaf4174cd279c7ecc86bbf4e3d1/vendor/pipenv-to-pip#L10

The json library in python3 versions earlier than 3.6 does not play well with binary input.
Trying to `json.load()` a file opened in binary mode leads to a `TypeError`:
```
$ python3
Python 3.4.3 (default, Sep 12 2017, 16:23:29)
>>> import json
>>> with open('Pipfile.lock', 'rb') as infile:
...    x = json.load(infile)
...
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
  File "<edited>/python/lib64/python3.4/json/__init__.py", line 268, in load
    parse_constant=parse_constant, object_pairs_hook=object_pairs_hook, **kw)
  File "<edited>/python/lib64/python3.4/json/__init__.py", line 312, in loads
    s.__class__.__name__))
TypeError: the JSON object must be str, not 'bytes'
>>>
```

This is causing our heroku build to fail with the above traceback.
Switching to `r` mode instead of `rb` should be an easy fix to this problem (assuming that doesn't conflict with anything else)

kennethreitz:
pull requests accepted! Note, we officially only endorse the use of latest 2.7.x and 3.6.x.

PR #688, opened 2018-04-25
Use pipenv lock instead of pipenv-to-pip

dz0ny:
Ref: https://github.com/heroku/heroku-buildpack-python/issues/678


kennethreitz:
Lock is not designed to work with --system. It creates a virtualenv, which is expensive, both in time and disk space.

PR #698, opened 2018-05-07
Update pipenv-to-pip to read file in `r` mode

pdamodara:
I tested `pipenv-to-pip` locally, seems to work as expected.

```
# On branch master
$ python3.4 vendor/pipenv-to-pip <path_to_Pipfile.lock>
Traceback (most recent call last):
  File "vendor/pipenv-to-pip", line 23, in <module>
    main()
  File "vendor/pipenv-to-pip", line 11, in main
    lockfile = json.load(f)
  File "/home/fds/lang/python/lib64/python3.4/json/__init__.py", line 268, in load
    parse_constant=parse_constant, object_pairs_hook=object_pairs_hook, **kw)
  File "/home/fds/lang/python/lib64/python3.4/json/__init__.py", line 312, in loads
    s.__class__.__name__))
TypeError: the JSON object must be str, not 'bytes'

# On branch fix-pipenv-to-pip
$ python3.4 vendor/pipenv-to-pip <path_to_Pipfile.lock>
# Prints out `requirements.txt`-style requirements
```

Also tested that it still works for `python2.7` and `python3.6`.

### DIFF
--- a/vendor/pipenv-to-pip
+++ b/vendor/pipenv-to-pip
@@ -7,7 +7,7 @@ import sys
 def main():
     INFILE = sys.argv[1]
 
-    with open(INFILE, 'rb') as f:
+    with open(INFILE, 'r') as f:
         lockfile = json.load(f)
 
     packages = []


vendor/pipenv-to-pip
```diff
@@ -7,7 +7,7 @@ import sys
 def main():
     INFILE = sys.argv[1]
 
-    with open(INFILE, 'rb') as f:
+    with open(INFILE, 'r') as f:
```
sigmavirus24: This is an excellent catch. I think a slightly better solution might be to use `io.open` with an explicit encoding to turn it into text that `json.load` could then parse. It would look like:

```py
import io
# ...

def main():
    INFILE = sys.argv[1]

    with io.open(INFILE, 'r', encoding='utf-8') as f:
        lockfile = json.load(f)

# ...
```

pdamodara: Agreed. This will ensure that `2.7.x` compatibility stays intact.

pdamodara:
The build failed because it timed out. I'm... not sure how to fix that.

sigmavirus24:
> The build failed because it timed out. I'm... not sure how to fix that.

If you update the code with my suggestion, it should start a new build.

PR #705, opened 2018-06-01
Fix incorrenct pip version in pip-uninstall step

### LINKED ISSUE #700: ModuleNotFoundError: No module named 'pip.req'  during installing pip
Following error occurs during installing pip

```
-----> Installing pip        
remote: Traceback (most recent call last):        
remote:   File "/tmp/buildpackKXfaB/vendor/pip-pop/pip-diff", line 15, in <module>        
remote:     from pip.req import parse_requirements        
remote: ModuleNotFoundError: No module named 'pip.req'        
remote: mount: can't find failure.bad-requirements in /etc/fstab or /etc/mtab 
```

This error occurs during every new rebulding.

jxltom:
Try to fix #700 

Currently pip version is pinned to ```9.0.2``` and is installed in ```/steps/python```

Then pipenv is installed in ```steps/pipenv```. This step will also upgrade pipenv which means pip is changed back to latest version 10.

Finally ```steps/pip-uninstall``` is called and this step will call ```pip-diff``` which called ```from pip.req import parse_requirements```. However, pip 10 does not has ```pip.req``` and then exception is raised.

So ```--upgrade-strategy only-if-needed``` is added when installing ```pipenv``` to make sure only ```pipenv``` is upgraded while ```pip``` is kept as older version.

### DIFF
--- a/.idea/heroku-buildpack-python.iml
+++ b/.idea/heroku-buildpack-python.iml
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module type="PYTHON_MODULE" version="4">
+  <component name="NewModuleRootManager">
+    <content url="file://$MODULE_DIR$" />
+    <orderEntry type="inheritedJdk" />
+    <orderEntry type="sourceFolder" forTests="false" />
+  </component>
+  <component name="TestRunnerService">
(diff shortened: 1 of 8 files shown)


jxltom:
Close since latest version of pipenv requires pip 10. The ```pip-diff``` should support pip 10 instead.

Issue #702, opened 2018-05-25
Deploy fails when uploading Pipfile.lock


ddahan:
When I deploy my project with Pipfile and Pipfile.lock, here is the error I got on Heroku:

```
-----> Python app detected
-----> Installing dependencies with Pipenv 11.8.2…
       Your Pipfile.lock (48bdc8) is out of date. Expected: (2bfadf).
       Aborting deploy.
 !     Push rejected, failed to compile Python app.
 !     Push failed
```

Note that my Pipfile.lock is up to date since I ran `pipenv update` just before commiting, pushing and deploying.

If I erase my Pipfile.lock, then commiting, pushing, deploying, it works.
However, it's not a clean solution because the Pipfile.lock must be updated to ensure deterministic builds.

I'm using Python 3.6.5 and Pipenv `2018.05.18`.
Not sure why the buildpack hardcoded the `11.8.2` version.

sigmavirus24:
Related to #701

CaseyFaist:
Buildpack is currently pinned at Pipenv version `2018.05.18` - are you still seeing this behavior @ddahan?

PR #706, opened 2018-06-01
Downgrade pipenv version to v11.9.0

### LINKED ISSUE #704: Deploy fails due to outdated Pipfile.lock
Our heroku builds have started failing today with the error message
```
Your Pipfile.lock (<local hash>) is out of date. Expected: (<new hash>)
```
I notice that you have bumped the pipenv version today, and I have also updated my local pip (`10.0.1`) and and pipenv (`2018.05.18`) installations. However, I am unable to get my Pipfile.lock hash to change to `<new hash>` -running `pipenv lock` does not change the hash at all.

I am at a loss as to what is causing my local Pipfile.lock hash to be different to the one generated in the heroku build, or how to fix the problem.

jxltom:
Try to fix #704 #702

Pipenv with version newer than v11.9.0 has bug related with editable packages. It may related to https://github.com/pypa/pipenv/issues/2219. So downgrade it to versin v11.9.0.

When pipenv has fixed this bug, the version can be upgraded to latest version.

### DIFF
--- a/bin/steps/pipenv
+++ b/bin/steps/pipenv
@@ -41,7 +41,7 @@ if [ ! "$SKIP_PIPENV_INSTALL" ]; then
             export PIP_EXTRA_INDEX_URL
         fi
 
-        export PIPENV_VERSION="2018.5.18"
+        export PIPENV_VERSION="11.9.0"
 
         # Install pipenv.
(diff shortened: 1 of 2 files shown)


jxltom:
Close in favour of https://github.com/heroku/heroku-buildpack-python/pull/727

PR #709, opened 2018-06-08
Build libgdal with libgeos support

sigmavirus24:
According to https://www.pythonanywhere.com/forums/topic/2801/ this will
allow users of osgeo, geopandas, geodjango, and other related libraries
to operate without seeing a message like `ERROR 6: GEOS support not
enabled.` in their application logs.


CaseyFaist:
From issue #852 - GDAL support in the buildpack will be deprecated (with a grace period) in favor of a language-agnostic third party buildpack 👍

PR #711, opened 2018-06-11
Update puts-step "Installing pip"

JanNash:
Since the output of the actual pip install is redirected into /dev/null, one never knows (by just reading the deploy output) which pip version is installed. Adding the version to the puts-step should make for quicker debugging. It could be considered to also echo the version [in case it's already installed](https://github.com/heroku/heroku-buildpack-python/pull/711/files#diff-57a28f8b2bf70b3efa895ae239e9d986L62).

---------

#### On another note: 
Personally, I'd also be in favour of making it possible to configure the pip version that should be installed via a config variable. If this is wanted, I'll gladly work on it and submit a PR.

### DIFF
--- a/bin/steps/python
+++ b/bin/steps/python
@@ -61,7 +61,7 @@ fi
 # If Pip isn't up to date:
 if [ "$FRESH_PYTHON" ] || [[ ! $(pip --version) == *$PIP_UPDATE* ]]; then
 
-  puts-step "Installing pip"
+  puts-step "Installing pip $PIP_UPDATE"
 
   # Remove old installations.
   rm -fr /app/.heroku/python/lib/python*/site-packages/pip-*


JanNash:
Thanks for approving :)
Should I also add the change to echo the version in case it's already installed?

JanNash:
If there's any further input I can give, I'll gladly help :)

sigmavirus24:
> Should I also add the change to echo the version in case it's already installed?

@JanNash if you could rebase and add that, I'll merge this as soon as I notice it.

JanNash:
Hey, sorry for letting this lay stale for so long. Instead of rebasing, I re-forked and [created a new PR](https://github.com/heroku/heroku-buildpack-python/pull/792), so I'm closing this one now.

Issue #427, opened 2017-07-18
Issues with the vendored sqlite/libsqlite


edmorley:
Some context:
* Python has a run-time dependency on `libsqlite3-0` and a build-time dependency on `libsqlite3-dev`.
* The [Python build scripts](https://github.com/heroku/heroku-buildpack-python/blob/78767e71996ddde2de5fdb84fc1e0291cc5fe4e3/builds/runtimes/python-2.7.13#L3) used for this buildpack vendor a [custom-built sqlite](https://github.com/heroku/heroku-buildpack-python/blob/78767e71996ddde2de5fdb84fc1e0291cc5fe4e3/builds/libraries/sqlite) into `.heroku/python/`.
* The vendored sqlite archive includes three components: the shared library (a la `libsqlite3-0`), headers/static library (a la `libsqlite3-dev`), and the sqlite3 CLI binary. 
* The vendored archive is used whilst the Python binary is being compiled (otherwise it would skip the `_sqlite3` module) and then also included in the Python archive uploaded to S3 ready for the slug compile.

There are a few problems with this:
* The vendored sqlite is extremely old (v3.7.9, [released on 2011-11-01](https://www.sqlite.org/releaselog/3_7_9.html)). Compare this to [v3.11.0 on Xenial](https://packages.ubuntu.com/xenial/libsqlite3-0) and [v3.8.2 on Trusty](https://packages.ubuntu.com/trusty-updates/libsqlite3-0).
* Even if this version were updated now, keeping custom builds rather than using OS packages means:
  - an going battle to keep the version up to date with new releases/security fixes.
  - reduced parity with the Ubuntu Python and [Docker Python images]((https://github.com/docker-library/python/blob/c9954b06c8b178d7888bc1626bed5a14e43a9203/2.7/slim/Dockerfile#L14)) (that use the OS packages).
* It adds 12MB to the slug size on disk for every Python app (13% of the Python 2.7.13 archive size), even though part of the contents (`libsqlite3-0`) actually duplicates what is [installed in the stack image](https://devcenter.heroku.com/articles/stack-packages).

As such, I propose switching from vendoring sqlite to using the OS packages. I'm guessing the cedar-14 images will want to be left as-is, and so this only be done for Heroku-16.

This would require:
1) Adding `libsqlite3-dev` to the `heroku:16-build` image.
[...]

PR #713, opened 2018-06-21
Update SQLite3

hone:
With inspiration from @KevinBrolly, this patch uses the stack image SQLite3 package but also still providing the dev headers and binary that users may still be using today. We can just stop shipping SQLite3 from future binaries. In addition, we don't need to worry about what version and when to update SQLite3 and maintaining the packages ourselves.

I limited the patches to Python 2.7.15 and 3.6.6 with this change, so we wouldn't have to rebuild all the binaries. We'll need re-release the binaries in conjunction with this buildpack change. The binary building and buildpack share the same sqlite3 installation code except the binary building only installs the dev headers for python to build with the stack image.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -160,6 +160,8 @@ cp -R "$CACHE_DIR/.heroku/python" .heroku/ &> /dev/null || true
 cp -R "$CACHE_DIR/.heroku/python-stack" .heroku/ &> /dev/null || true
 # A plain text file which contains the current python version being used (used for cache busting).
 cp -R "$CACHE_DIR/.heroku/python-version" .heroku/ &> /dev/null || true
+# A plain text file which contains the current sqlite3 version being used (used for cache busting).
+cp -R "$CACHE_DIR/.heroku/python-sqlite3-version" .heroku/ &> /dev/null || true
 # Any pre-compiled binaries, provided by the buildpack.
 cp -R "$CACHE_DIR/.heroku/vendor" .heroku/ &> /dev/null || true
 # "editable" installations of code repositories, via pip or pipenv.
@@ -273,6 +275,16 @@ sub_env "$BIN_DIR/steps/geo-libs"
 # This is part of the Geo support.
 # shellcheck source=bin/steps/gdal
 source "$BIN_DIR/steps/gdal"
+
+# SQLite3 support.
+# This sets up and installs sqlite3 dev headers and the sqlite3 binary but not the
+# libsqlite3-0 library since that exists on the stack image.
+# Note: This only applies to Python 2.7.15+ and Python 3.6.6+
+(( start=$(nowms) ))
+# shellcheck source=bin/steps/sqlite3
+source "$BIN_DIR/steps/sqlite3"
+buildpack_sqlite3_install
+mtime "sqlite3.install.time" "${start}"
 
 # pip install
 # -----------
--- a/bin/steps/python
+++ b/bin/steps/python
@@ -29,7 +29,15 @@ fi
 
 if [[ "$STACK" != "$CACHED_PYTHON_STACK" ]]; then
     puts-step "Stack has changed from $CACHED_PYTHON_STACK to $STACK, clearing cache"
-    rm -fr .heroku/python-stack .heroku/python-version .heroku/python .heroku/vendor
+    rm -fr .heroku/python-stack .heroku/python-version .heroku/python .heroku/vendor .heroku/python .heroku/python-sqlite3-version
+fi
+
+# need to clear the cache for first time installing SQLite3,
+# since the version is changing and could lead to runtime errors
+# with compiled extensions.
+if [ -d .heroku/python ] && [ ! -f .heroku/python-sqlite3-version ] && python_sqlite3_check "$PYTHON_VERSION"; then
+  puts-step "Need to update SQLite3, clearing cache"
+  rm -fr .heroku/python-stack .heroku/python-version .heroku/python .heroku/vendor
 fi
 
 if [ -f .heroku/python-version ]; then
--- a/bin/steps/sqlite3
+++ b/bin/steps/sqlite3
@@ -1,0 +1,87 @@
+#!/usr/bin/env bash
+
+# shellcheck source=bin/utils
+source "$BIN_DIR/utils"
+
+sqlite3_version() {
+  SQLITE3_VERSION=${SQLITE3_VERSION:-$(dpkg -s libsqlite3-0 | grep Version | sed 's/Version: //')}
+
+  export SQLITE3_VERSION
+}
+
+sqlite3_install() {
+  HEROKU_PYTHON_DIR="$1"
+  SQLITE3_VERSION="$2"
+  HEADERS_ONLY="$3"
+
+  mkdir -p "$HEROKU_PYTHON_DIR"
+
+  APT_CACHE_DIR="$HEROKU_PYTHON_DIR/apt/cache"
+  APT_STATE_DIR="$HEROKU_PYTHON_DIR/apt/state"
+
+  mkdir -p "$APT_CACHE_DIR/archives/partial"
+  mkdir -p "$APT_STATE_DIR/lists/partial"
+
+  APT_OPTIONS="-o debug::nolocking=true"
+  APT_OPTIONS="$APT_OPTIONS -o dir::cache=$APT_CACHE_DIR"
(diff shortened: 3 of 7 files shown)


bin/steps/python
```diff
@@ -27,6 +27,14 @@ if [[ "$STACK" != "$CACHED_PYTHON_STACK" ]]; then
     rm -fr .heroku/python-stack .heroku/python-version .heroku/python .heroku/vendor
```
edmorley: I think `.heroku/python-sqlite3-version` needs to be added to the list of files deleted here, since otherwise if the stack changes, the sqlite binary and headers end up being deleted, but yet the `bin/steps/sqlite3` won't reinstall them, since `.heroku/python-sqlite3-version` will still exist.

Alternatively, perhaps the sqlite3 version file could be moved into `.heroku/python/`, since they are quite strongly tied together?

bin/steps/sqlite3
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env bash
+
+set +e
```
edmorley: This steps file is sourced, so the `set +e` here will affect subsequent steps too. This may be fine, but mentioning it just in case.

bin/steps/sqlite3
```diff
@@ -0,0 +1,64 @@
...
+
+  puts-step "Installing SQLite3"
+  apt-get $APT_OPTIONS update >/dev/null 2>&1
+  apt-get $APT_OPTIONS -y -d --reinstall install libsqlite3-dev sqlite3 >/dev/null 2>&1
+
+  for deb in $(find $APT_CACHE_DIR/archives/ -name "*.deb"); do
+    dpkg -x $deb "$BUILD_DIR/.heroku/python/sqlite3/"
+  done
+
+  HEROKU_PYTHON_DIR="$BUILD_DIR/.heroku/python"
```
edmorley: The Python directory is referenced manually a few times above - perhaps it would be worth moving `HEROKU_PYTHON_DIR` up higher and using for those instances too?

bin/steps/sqlite3
```diff
@@ -0,0 +1,64 @@
...
+
+SQLITE3_VERSION_FILE="$BUILD_DIR/.heroku/python-sqlite3-version"
+if [ -f $SQLITE3_VERSION_FILE ]; then
+  INSTALLED_SQLITE3_VERSION=$(cat $SQLITE3_VERSION_FILE)
+fi
+
+# only install if the sqlite3 version has changed
+if [[ $INSTALLED_SQLITE3_VERSION != $SQLITE3_VERSION ]]; then
+  APT_CACHE_DIR="$CACHE_DIR/.heroku/python/apt/cache"
+  APT_STATE_DIR="$CACHE_DIR/.heroku/python/apt/state"
```
edmorley: I wonder if at some point in the future it would be worth adding some APT helper functions to buildpack-stdlib to reduce the duplication between the various buildpacks that have to play the non-system APT install dance?

bin/compile
```diff
@@ -273,6 +275,13 @@ sub_env "$BIN_DIR/steps/geo-libs"
 # shellcheck source=bin/steps/gdal
 source "$BIN_DIR/steps/gdal"
 
+# SQLite3 support.
+# This sets up and installs sqlite3 dev headers and the sqlite3 binary
```
edmorley: Is it worth mentioning here or in `bin/steps/sqlite3` that:
* `libsqlite3-0` isn't being installed here since it already exists in the stack images
* the reason the headers aren't installed in the stack image is https://github.com/heroku/heroku-buildpack-python/issues/427#issuecomment-316794743

bin/steps/python
```diff
@@ -27,6 +27,14 @@ if [[ "$STACK" != "$CACHED_PYTHON_STACK" ]]; then
...
 fi
 
+# need to clear the cache for first time installing SQLite3,
+# since the version is changing and could lead to runtime errors
+# with compiled extensions.
+if [ -d .heroku/python ] && [ ! -f .heroku/python-sqlite3-version ]; then
+  puts-step "Need to update SQLite3, clearing cache"
+  rm -fr .heroku/python-stack .heroku/python-version .heroku/python .heroku/vendor
```
sigmavirus24: It seems we're duplicating this and we may be missing files, should we have an array of files we want to delete or a single `clear-cache` function to use?

bin/steps/sqlite3
```diff
@@ -0,0 +1,51 @@
...
+apt-get $APT_OPTIONS -y -d --reinstall install libsqlite3-dev sqlite3 >/dev/null 2>&1
+
+for deb in $(find $APT_CACHE_DIR/archives/ -name "*.deb"); do
+  dpkg -x $deb "$BUILD_DIR/.heroku/python/sqlite3/"
+done
+
+HEROKU_PYTHON_DIR="$BUILD_DIR/.heroku/python"
+# remove old sqlite3 libraries/binaries
+for header in $(find $HEROKU_PYTHON_DIR/include/ -name "sqlite3*.h"); do
+  rm -f $header
```
schneems: Is it possible that one of the headers will have a space? Should we quote "$header" ?

bin/steps/sqlite3
```diff
@@ -0,0 +1,51 @@
...
+done
+for lib in $(find $HEROKU_PYTHON_DIR/lib/ -name "libsqlite3.*"); do
+  rm -f $lib
```
schneems: Same question here. Should this be quoted?

bin/steps/sqlite3
```diff
@@ -0,0 +1,64 @@
...
+
+SQLITE3_VERSION=$(apt-cache policy libsqlite3-dev | grep Candidate | sed 's/  Candidate: //')
+
+SQLITE3_VERSION_FILE="$BUILD_DIR/.heroku/python-sqlite3-version"
+if [ -f $SQLITE3_VERSION_FILE ]; then
+  INSTALLED_SQLITE3_VERSION=$(cat $SQLITE3_VERSION_FILE)
+fi
+
+# only install if the sqlite3 version has changed
+if [[ $INSTALLED_SQLITE3_VERSION != $SQLITE3_VERSION ]]; then
```
KevinBrolly: I am not sure that this will ever be true, the cache is blank, so `$INSTALLED_SQLITE3_VERSION` is null, `apt-cache policy libsqlite3-dev | grep Candidate | sed 's/  Candidate: //'` will only return if `apt-get update` is run first, so I am not sure if this will work in the current context of the buildpack? If it is the case that `apt-cache policy libsqlite3-dev | grep Candidate | sed 's/  Candidate: //'` returns null then this condition will not be true unless the `$BUILD_DIR/.heroku/python-sqlite3-version` file exists, which it will not because it is only created at the end of this conditional block. Can anyone confirm this?

bin/steps/python
hone: Good catch. I kept it out to keep it in line with `.heroku/python-version`.

bin/compile
```diff
@@ -274,6 +276,16 @@ sub_env "$BIN_DIR/steps/geo-libs"
 # shellcheck source=bin/steps/gdal
 source "$BIN_DIR/steps/gdal"
 
+# SQLite3 support.
+# This sets up and installs sqlite3 dev headers and the sqlite3 binary but not the
+# libsqlite3-0 library since that exists on the stack image.
+# Note: This only applies to Python 2.7.15+ and Python 3.6.5+
```
KevinBrolly: You said `Python 3.6.5+` here but the runtime build script for 3.6.5 has not been updated to install Sqlite3

hone: Good catch! Thanks.

bin/compile
```diff
@@ -274,6 +276,16 @@ sub_env "$BIN_DIR/steps/geo-libs"
...
+(( start=$(nowms) ))
+# shellcheck source=bin/steps/sqlite3
+source "$BIN_DIR/steps/sqlite3"
```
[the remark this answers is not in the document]
hone: We need to use it in both since the dev headers are missing on the stack image, so it won't be around for build or runtime. We need it at build time, so the python sqlite library is built against the version of sqlite3 on the stack image.

hone:
@KevinBrolly you're right, I'll update the PR.

builds/runtimes/python-3.6.6
```diff
@@ -3,6 +3,15 @@
 # Build Deps: libraries/sqlite
```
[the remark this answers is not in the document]
hone: yes, but not in 3.6.5. fixing.

PR #714, opened 2018-06-22
allow to set pip version

pedrocarmona:
Hello everyone,

I am having a hard time installing airflow, turns out the latest pip just wont work. In the meanwhile, I am using pip 9.0.3. With this PR its possible to set pip to 9.0.3 as an compile time env variable.

Thank you all for you time :)

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -56,7 +56,8 @@ LATEST_2="python-2.7.15"
 # Which stack is used (for binary downloading), if none is provided (e.g. outside of Heroku)?
 DEFAULT_PYTHON_STACK="cedar-14"
 # If pip doesn't match this version (the version we install), run the installer.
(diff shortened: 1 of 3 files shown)


CaseyFaist:
Hey there @pedrocarmona - Thanks for the pr! A pip update is being considered, however I'm curious if your issue isn't related to Airflow's package name on PyPI changing from `airflow` to `apache-airflow` when it went into Apache Incubation (documentation [here](https://pypi.org/project/apache-airflow/)). I understand users may need the older version if working with a legacy codebase, however if you're starting a new project, I recommend using the latest stable release. 

If you need a legacy version - 1.8.0 or earlier - I was also able to confirm that if you specify the version in your requirements.txt file or in your pip command, Airflow installs successfully - so an Airflow installation is not dependent on the pip version used.

CaseyFaist:
Closing, as issue is about package naming, and not with pip version.

PR #715, opened 2018-06-25
Make the profile.d script sh-compatible

dmathieu:
The `[[` statement is a bashism. Bash may not be always available (especially in docker images).
Sh should always be.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -330,7 +330,7 @@ set_default_env PYTHONPATH "\$HOME"
 # Python expects to be in /app, if at runtime, it is not, set
 # up symlinks… this can occur when the subdir buildpack is used.
 cat <<EOT >> "$PROFILE_PATH"
-if [[ \$HOME != "/app" ]]; then
+if [ \$HOME != "/app" ]; then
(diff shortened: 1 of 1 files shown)


hone:
:+1: to the sentiment above

dmathieu:
The other bashisms should be fine, as they happen at build time.
This is executed at runtime where, in docker environments, bash isn't necessarily the default shell.

hunterloftis:
Right now, bash *is* necessarily the default shell because that's a guarantee that has been made to Buildpack authors, through the Buildpack API docs, for many years: "Like the standard Linux /etc/profile.d shell startup files, these must be bash scripts."

We could evolve the Buildpack API to be strictly sh-compatible, but that should be an explicit choice vs something we special-case into the buildpacks that we control on a 1-off basis when we need to.

dmathieu:
All right, I'll wait until my bug in runtime is fixed.

hunterloftis:
Instead of changing & publishing the official buildpack, I'd recommend using a fork for testing (if you're currently being blocked by an sh-vs-bash bug).

PR #716, opened 2018-06-25
SQLite3 Upgrade

KevinBrolly:
Hey @hone after looking at your PR I wanted to just put something together to suggest a slightly different way we could tackle this issue.

This updates all the build scripts to just install the needed `libsqlite3-dev` library when building python to enable support for sqlite3 in the standard library, rather than also installing the binary CLI and `libsqlite3` that is already available in the stack image.

It then also installs the sqlite CLI binary as a step in the build process at a set version and caches it, and moves it to the place that the current sqlite3 binary is located in the build, so it should not break for anyone currently using this functionality.

I then added another commit adding a `BUILD_WITH_SQLITE` environment variable to make the sqlite3 binary an opt-in feature rather than automatic, as I really don't think that there are a lot of users who would need this (ephemeral filesystem). I also added an mcount call to track the number of builds who opt-in to installing the sqlite3 CLI binary, this could provide a deprecation path in the future if there are not many people enabling this.

If we did not incorporate the last commit with the "opt-in" then this would work exactly as the buildpack works now, except there is not trace of the old sqlite3 version, and we can easily update the version in the future.

I have a test app running this buildpack (with the rebuilt python runtimes) and have tested it with 3.6.5 and 2.7.15


builds/runtimes/pypy-5.3.1
```diff
@@ -1,11 +1,13 @@
 #!/usr/bin/env bash
 # Build Path: /app/.heroku/python/
-# Build Deps: libraries/sqlite
 
 # NOTICE: This formula only works for the cedar-14 stack, not cedar.
 
 OUT_PREFIX=$1
 
+echo "Installing libsqlite3-dev"
+apt-get -qq update && apt-get install libsqlite3-dev
```
hone: don't you need super user permission here?

hone:
Nice! I had similar thoughts and have been working on a patch to install the dev headers with the build and couple that with my other PR. I also question how much people are using SQLite3 as an ephemeral store, but at this point don't want to rock the boat and leave that to the new Python person when they come in.

builds/runtimes/pypy-5.3.1
hone: It looks like the cedar-14 builder doesn't have super user since it's not using the container runtime.

sigmavirus24:
@hone @KevinBrolly could you two come together on a PR to fix the sqlite3dev issue here? I don't have the time to tackle it myself but can lend pull request reviews.

hone:
@sigmavirus24 I updated my PR #713

PR #722, opened 2018-06-28
Build Python 3.6.6 and 3.7.0

sigmavirus24:
Add runtime build recipes for our newly released versions

Closes gh-720
Closes gh-721


hone:
@sigmavirus24 should we set the default 3.6 to 3.6.6?

PR #725, opened 2018-06-30
mssql support

GitBib:


sigmavirus24:
I don't believe this actually is of any assistance to you for achieving mssql support on the Heroku platform. This dockerfile is only used to build the version of Python which does not rely on the presence of mssql-tools to provide support to the language.

PR #727, opened 2018-07-06
Update pipenv version

harrywhite4:
An issue with pipenv's --system --deploy options has been fixed in a newer release as described here
https://github.com/pypa/pipenv/blob/master/HISTORY.txt#L95

This was likely the casue of #704 and #702 so it would be nice to switch to the latest version

### DIFF
--- a/bin/steps/pipenv
+++ b/bin/steps/pipenv
@@ -41,7 +41,7 @@ if [ ! "$SKIP_PIPENV_INSTALL" ]; then
             export PIP_EXTRA_INDEX_URL
         fi
 
-        export PIPENV_VERSION="2018.5.18"
+        export PIPENV_VERSION="2018.7.1"
 
         # Install pipenv.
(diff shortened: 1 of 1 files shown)


harrywhite4:
I understand it's difficult to find a version that everyone will be happy with. Being able to set an environment variable to override the version might be a better solution

sigmavirus24:
Perhaps an `OVERRIDDEN_PIPENV_VERSION` that overrides our default?

techalchemy:
@sigmavirus24 if you file issues with us or you let me know directly about things that impact your pipeline I can triage these more effectively -- my company has an enterprise account with heroku and these do have an impact on me too so I'm happy to help

sigmavirus24:
@techalchemy I referenced exactly the issues that would be blockers above. It sounds like you're expecting something else though, what do you think I should have done differently?

harrywhite4:
Closing in favor of #732

Issue #728, opened 2018-07-06
Using python 3.7.0 raises warning "This version is unsupported"


KevinBrolly:
Currently we are supporting:

2.7.15
3.6.6 (default)
3.7.0

The buildpack does not currently have the ability to support 2 separate minor versions (3.6.6 and 3.7.0 in this case) without outputting a "(you are using python-x.x.x, which is unsupported)" message.

This is because we only have a single variable for $LATEST_3 - https://github.com/heroku/heroku-buildpack-python/blob/a0275888a26b8f0132c7bb9fe3bf194b5c4fab1f/bin/steps/python#L18

DEKHTIARJonathan:
I still have the same issue.

### With python-3.6.6

```
Staging...
-----> Installing python-3.6.6
 !     Requested runtime (python-3.6.6) is not available for this stack (cedar-14).
 !     Aborting.  More info: https://devcenter.heroku.com/articles/python-support
Failed to compile droplet: Failed to compile droplet: exit status 1
```

### With python-3.7.0

```
Staging...
Downloaded build artifacts cache (602.1M)
 !     The latest version of Python 3 is python-3.6.6 (you are using python-3.7.0, which is unsupported).
 !     We recommend upgrading by specifying the latest version (python-3.6.6).
       Learn More: https://devcenter.heroku.com/articles/python-runtimes
-----> Installing python-3.7.0
 !     Requested runtime (python-3.7.0) is not available for this stack (cedar-14).
 !     Aborting.  More info: https://devcenter.heroku.com/articles/python-support
Failed to compile droplet: Failed to compile droplet: exit status 1
```

iocohen:
me too!

agstew:
Requested runtime (Python 3.7.0) is not available for this stack (heroku-18)
Got the problem too!

WisnuBramantyo:
I got the same issue with Python 3.6.6 on heroku-18.
`remote: -----> Python app detected`
`remote: -----> Installing Python 3.6.6`
`remote:  !     Requested runtime (Python 3.6.6) is not available for this stack (heroku-18).`
`remote:  !     Aborting.  More info: https://devcenter.heroku.com/articles/python-support`
`remote:  !     Push rejected, failed to compile Python app.`
What version of python we should use actually?

hiroakikitaji:
same issue with Python 3.7.0

`remote: -----> Python app detected
[...]

PR #729, opened 2018-07-06
Prevent Python 3.7 from being unsupported

sigmavirus24:
Python 3.7.0 is supported but not preferred given how new it is. As a
result, we don't want it to be the default, but we also don't want users
to be confused when upgrading to it.

Closes gh-728


bin/steps/python
```diff
@@ -14,12 +14,17 @@ if [[ $PYTHON_VERSION =~ ^python-2 ]]; then
     echo "       Learn More: https://devcenter.heroku.com/articles/python-runtimes"
   fi
 else
-  if [[ "$PYTHON_VERSION" != "$LATEST_3" ]]; then
-    puts-warn "The latest version of Python 3 is $LATEST_3 (you are using $PYTHON_VERSION, which is unsupported)."
-    puts-warn "We recommend upgrading by specifying the latest version ($LATEST_3)."
+  if [[ $PYTHON_VERSION =~ ^python-3.7 ]] && [[ "$PYTHON_VERSION" != "$LATEST_37" ]]; then
+    puts-warn "The latest version of Python 3 is $LATEST_37 (you are using $PYTHON_VERSION, which is unsupported)."
```
edmorley: Should this say "of Python 3.7" (and similarly for 3.6 below)? :-)

DEKHTIARJonathan:
@sigmavirus24 @KevinBrolly I keep getting the same error.

See : https://github.com/heroku/heroku-buildpack-python/issues/728#issuecomment-403751189

KevinBrolly:
@DEKHTIARJonathan You are actually seeing two separate problems, the first one is solved by this PR:

```
!     The latest version of Python 3 is python-3.6.6 (you are using python-3.7.0, which is unsupported).
!     We recommend upgrading by specifying the latest version (python-3.6.6).
      Learn More: https://devcenter.heroku.com/articles/python-runtimes
```

When this PR is released the message will correctly indicate that you are on the latest version.

The second problem is this:

```
 !     Requested runtime (python-3.7.0) is not available for this stack (cedar-14).
```

That however is a bit of a head-scratcher as python-3.7.0 is available for cedar-14, I have checked this myself and you should not see this error.  I noticed at the bottom of your error message the following:

`Failed to compile droplet: Failed to compile droplet: exit status 1`

Are you running the buildpack on DigitalOcean? The only thing I can think of that could be the problem is that DigitalOcean may be using a different `BUILDPACK_VENDOR_URL` on their platform that does not have python-3.7.0, could you perhaps open a support request with them and check?

DEKHTIARJonathan:
I'm actually using +/- raw CF on IBM Bluemix and I use Python 3.6.6 on production.

```yaml
defaults: &defaults
  domain: eu-gb.mybluemix.net
  timeout: 120
  memory: 768M
  disk_quota: 2G
  path: .
  stack: cflinuxfs2
  # buildpack: https://github.com/heroku/heroku-buildpack-python.git  # does not work
  buildpack: https://github.com/cloudfoundry/buildpack-python.git     # works perfectly
```

runtime.txt
```
python-3.6.6
```

sigmavirus24:
So it sounds like you're having issues using a different stack image (which likely sets a different STACK variable) with our buildpack and want us to be completely compatible with a moving target (the cloudfoundry image/buildpack). Am I missing something here?

DEKHTIARJonathan:
Oh maybe I missed something, I thought you were using the same stack.
Which stack shall I use with this BP ?

sigmavirus24:
You can use any of the three that we test against: https://travis-ci.org/heroku/heroku-buildpack-python/builds/402204669, e.g., `heroku/heroku:18-build`. The full list is here: https://hub.docker.com/r/heroku/heroku/

DEKHTIARJonathan:
Oh my bad. I try this immediately. Thanks for the help

PR #741, opened 2018-08-17
Fix Python 3.6 message when using Python 3.7

CaseyFaist:
Was still seeing Python 3.6 error when creating 3.7 apps, so dug into the version nesting a bit.

Added nested layer for bucketing python 3 installs, then buckets for 3.6 and 3.7 independently.

Added catch for python 3 versions below 3.6 - these are not available on default stack, but deploy log now explicitly states that 3.4 and 3.5, etc, are not supported.

### DIFF
--- a/bin/steps/python
+++ b/bin/steps/python
@@ -12,17 +12,29 @@ if [[ $PYTHON_VERSION =~ ^python-2 ]]; then
     puts-warn "The latest version of Python 2 is $LATEST_2 (you are using $PYTHON_VERSION, which is unsupported)."
     puts-warn "We recommend upgrading by specifying the latest version ($LATEST_2)."
     echo "       Learn More: https://devcenter.heroku.com/articles/python-runtimes"
+  else
+    echo "       Using supported version of Python 2 ($PYTHON_VERSION)"
   fi
 else
-  if [[ $PYTHON_VERSION =~ ^python-3.7 ]] && [[ "$PYTHON_VERSION" != "$LATEST_37" ]]; then
-    puts-warn "The latest version of Python 3.7 is $LATEST_37 (you are using $PYTHON_VERSION, which is unsupported)."
-    puts-warn "We recommend upgrading by specifying the latest version ($LATEST_37)."
-    echo "       Learn More: https://devcenter.heroku.com/articles/python-runtimes"
-  else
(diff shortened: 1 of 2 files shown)


bin/steps/python
```diff
@@ -12,17 +12,34 @@ if [[ $PYTHON_VERSION =~ ^python-2 ]]; then
     puts-warn "The latest version of Python 2 is $LATEST_2 (you are using $PYTHON_VERSION, which is unsupported)."
     puts-warn "We recommend upgrading by specifying the latest version ($LATEST_2)."
     echo "       Learn More: https://devcenter.heroku.com/articles/python-runtimes"
+  else
+    echo "       Using supported version of Python 2 ($PYTHON_VERSION)"
```
hone: I believe this block is just for displaying warnings. Did you want to have this in addition to the rest of the buildpack output?

bin/steps/python
```diff
@@ -12,17 +12,34 @@ if [[ $PYTHON_VERSION =~ ^python-2 ]]; then
...
 else
-  if [[ $PYTHON_VERSION =~ ^python-3.7 ]] && [[ "$PYTHON_VERSION" != "$LATEST_37" ]]; then
-    puts-warn "The latest version of Python 3.7 is $LATEST_37 (you are using $PYTHON_VERSION, which is unsupported)."
-    puts-warn "We recommend upgrading by specifying the latest version ($LATEST_37)."
-    echo "       Learn More: https://devcenter.heroku.com/articles/python-runtimes"
-  else
-    if [[ "$PYTHON_VERSION" != "$LATEST_36" ]]; then
-      puts-warn "The latest version of Python 3.6 is $LATEST_36 (you are using $PYTHON_VERSION, which is unsupported)."
-      puts-warn "We recommend upgrading by specifying the latest version ($LATEST_36)."
+  if [[ $PYTHON_VERSION =~ ^python-3 ]]; then
```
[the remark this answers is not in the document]
CaseyFaist: 👏 yes, that's way nicer. updating, booya

bin/steps/python
CaseyFaist: I'd like to move the output in a direction where it's explicit (rather than silent) when a supported Python version is used. Open to chatting about other ways to go about that though - could be simplified to an operator for example

PR #742, opened 2018-08-21
Hatchet/yay

CaseyFaist:
This adds Hatchet integration testing for the Python Buildpack. It also:

- Updates Travis CI


schneems:
@CaseyFaist let me know when you've done the full account confirmation for the api user and then I can approve this and we can merge it in.

spec/hatchet/python_spec.rb
```diff
@@ -0,0 +1,10 @@
+require_relative '../spec_helper'
+
+describe "Python!!!!!!!!!!!" do
+  it "🐍" do
+    Hatchet::Runner.new('python-getting-started', stack: DEFAULT_STACK).deploy do |app|
```
jmorrell: @schneems can check me here, but I don't think you need to explicitly specify the `stack` here

spec/spec_helper.rb
```diff
@@ -0,0 +1,23 @@
+ENV['HATCHET_BUILDPACK_BASE'] = 'https://github.com/heroku/heroku-buildpack-python.git'
```
jmorrell: @schneems Does the hatchet rake task pick up the git branch?

spec/hatchet/python_spec.rb
schneems: Correct, but explicit > implicit for python. I think my intention here was to declare which tests should be stack independent so that they can all be changed with one variable change.

spec/spec_helper.rb
schneems: Git branch is auto detected on travis and heroku CI. If it's not detected on another CI provider let me know and I can add it. https://github.com/heroku/hatchet/blob/master/lib/hatchet/app.rb#L9 and https://github.com/heroku/hatchet/blob/master/lib/hatchet.rb#L28

spec/spec_helper.rb
```diff
@@ -0,0 +1,23 @@
...
+
+require 'rspec/core'
+require 'rspec/retry'
+require 'hatchet'
+
+require 'date'
```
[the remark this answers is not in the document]
schneems: This is due to a bug in hatchet https://github.com/heroku/hatchet/commit/94f7079e4c1ece2560003df8fe488c575c3dc85f.

spec/spec_helper.rb
```diff
@@ -0,0 +1,23 @@
...
+    c.syntax = :expect
+  end
+end
+
+if ENV['TRAVIS']
+  # Don't execute tests against "merge" commits
+  exit 0 if ENV['TRAVIS_PULL_REQUEST'] != 'false' && ENV['TRAVIS_BRANCH'] == 'master'
+end
+
+DEFAULT_STACK = 'heroku-16'
```
schneems: Newline here (eventually).

CaseyFaist: fixed, so I don't forget it

PR #743, opened 2018-08-22
Upgrade/rebase onto upstream

atzannes:
# Purpose
Merge in all the upstream changes that happened in the last 4 years because we have been having very regular buildpack breakages due to the code being so old.

Fixes [ALTOS-16888](https://altschool.atlassian.net/browse/ALTOS-16888)

# Approach
First tried to merge the upstream into our master. That rust failed.

Then tried to rebase our master onto upstream, which also lead to lots of extensive conflicts.

Because our changes were few, I reapplied them in separate commits that correspond to the changes we had made on top of that 4-yo upstream/master branch.

Because our fork was an indirect fork of the official upstream through Mikhail's fork, I modified our repo to be a direct fork of the upstream repo. Github does not allow an account to have multiple forks of the same repo, even if their source (Mikhail vs Heroku) are different. I tried renaming the repo in case the problem was a name conflict, but it wasn't, so I transferred ownership of our fork to my private account (https://github.com/atzannes/heroku-buildpack-python-old), created a fresh AltSchool fork of the Heroku repo, and cherry-picked all the relevant changes from the transferred repo to this one.

### DIFF
--- a/bin/steps/sshkey
+++ b/bin/steps/sshkey
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+
+# Generates an SSH config file for connections if a config var exists.
+
+# ENV_DIR=$3
+
+if [[ -f $ENV_DIR/CUSTOM_SSH_KEY && -f $ENV_DIR/CUSTOM_SSH_KEY_HOSTS ]]; then
+
+  #echo "" >&1
+  echo ""
+
+  # Ensure we have an ssh folder
+  if [ ! -d ~/.ssh ]; then
+    mkdir -p ~/.ssh
+    chmod 700 ~/.ssh
+  fi
+
+  # Load the private key into custom_key file.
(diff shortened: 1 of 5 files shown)


bin/steps/sshkey
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
+
+# Generates an SSH config file for connections if a config var exists.
```
edmorley: To reduce maintenance burden of your fork, you could move this into a separate buildpack, and use multiple buildpacks. In fact there is already a third-party SSH key buildpack available that you could use/fork:
https://elements.heroku.com/buildpacks/debitoor/ssh-private-key-buildpack
https://github.com/debitoor/ssh-private-key-buildpack

atzannes:
@edmorley thank you for the suggestion, it makes a lot of sense!

PR #745, opened 2018-08-24
update tests to expect success for all 2.7.15 builds

CaseyFaist:
This is a small fix on a unit test that expected a fail case on Python 2.7 on Heroku 18. 

Update to expect success for 2.7 builds on all three stacks.

### DIFF
--- a/test/run
+++ b/test/run
@@ -81,17 +81,9 @@ testPylibmc() {
 }
 
 testPython2() {
-  if [[ "$STACK" == "heroku-16" ]] || [[ "$STACK" == "cedar-14" ]]; then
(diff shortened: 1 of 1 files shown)


hone:
Congrats on shipping Python 2.7.15 for heroku-18!

test/run
```diff
@@ -81,18 +81,9 @@ testPylibmc() {
 }
 
 testPython2() {
-  if [[ "$STACK" == "heroku-16" ]] || [[ "$STACK" == "cedar-14" ]]; then
```
jmorrell: 👨‍🍳👌

PR #639, opened 2018-02-14
Add support for PIP_EXTRA_INDEX_URL

p-himik:
Akin to #363 but exports only `PIP_EXTRA_INDEX_URL` so users don't have to store possibly sensitive extra index URL in their repository code.

PR #750, opened 2018-09-17
Adding .python-compile-env for to support exporting env vars for pip install

tomasbielskis:
I added the ability to define config variables required for `pip install` to work for airflow 1.10. Airflow [now requires](https://github.com/apache/incubator-airflow/blob/master/UPDATING.md#airflow-110) the environment variable `SLUGIFY_USES_TEXT_UNIDECODE` to be defined and exported for app build in order for airflow to install.

By creating a file `.python-compile-env` in the app source root, it will be sourced before `pip install` runs. The file simply exports environment variables:
```
export SLUGIFY_USES_TEXT_UNIDECODE=yes
```

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -287,7 +287,7 @@ buildpack_sqlite3_install
 mtime "sqlite3.install.time" "${start}"
 
 # Source compile variables
-if [ -f .python-compile-env ]; then
+if [ -s .python-compile-env ]; then
   source .python-compile-env
 fi
 
--- a/bin/compile
+++ b/bin/compile
@@ -286,6 +286,11 @@ source "$BIN_DIR/steps/sqlite3"
 buildpack_sqlite3_install
 mtime "sqlite3.install.time" "${start}"
 
+# Source compile variables
+if [ -f .python-compile-env ]; then
+  source .python-compile-env
+fi
+
 # pip install
 # -----------
 


CaseyFaist:
Thanks for the PR! 

Now, I'm not sure adding library specific logic to the buildpack is the best route to this functionality. I think the underlying problem (environment variables created before `pip install`) has merit, but I'd sooner adopt something that can be adapted to solve similar problems for other use cases too.

Towards that goal, can you describe how this solves the Airflow problem in more detail? In particular:

- Why does setting the variable with the `os` library not work with Airflow in this case? 
- There's also dashboard tooling to control all environment variables documented at the link below. Are these incompatible with your use case?
https://devcenter.heroku.com/articles/config-vars#managing-config-vars

tomasbielskis:
Hi @CaseyFaist!

We tried using the config vars first, but they don't seem to be exposed to the buildpack in the environment itself. In fact, the [documentation](https://devcenter.heroku.com/articles/buildpack-api) suggests exactly that:

> The application config vars are passed to the buildpack as an argument (versus set in the environment) so that the buildpack can optionally export none, all or parts of the app config vars available in ENV_DIR when the compile script is run.

We can't set anything in the `os` library because that happens _after_ install and the environment variable need to be present _in order to_ install. 

We tried to solve this in the most proprietary software-agnostic way, if there's another way to accomplish this we're totally open to something better!

edmorley:
For previous art on this topic, see:
https://github.com/heroku/heroku-buildpack-python/pull/417
https://github.com/heroku/heroku-buildpack-python/pull/467
https://github.com/heroku/heroku-buildpack-python/pull/639

lmcgrath:
@edmorley I read the changes in those to see what I could glean. It looks like in #639 you did more or less what we changed (if file exists, read it into environment variable). The variable we're defining is proprietary (`SLUGIFY_USES_TEXT_UNIDECODE`) so we're being a little more willy-nilly by simply sourcing the file. This isn't ideal because the file can contain literally anything, but we haven't yet identified a better solution.

The bottom line is we can't use airflow 1.10 without this environment variable. There's no support in the buildpack to allow for arbitrary environment variables to be made available to `pip install`. I feel like this change is not palatable. Could you give us some specific feedback as to what we can do to make this acceptable?

edmorley:
I didn't write those PRs, nor do I have an opinion either way on this topic (and am not a Heroku employee); those links just seemed relevent to the discussion in case either of you hadn't seen them already :-)

lmcgrath:
I got you, thanks!

CaseyFaist:
Thanks for expanding @tomasbielskis and for more context @edmorley! I hadn't seen those PRs, I'm fairly new at Heroku myself :) I <3 airflow though, so I'm glad this issue was surfaced.

@lmcgrath not unpalatable - I just need to test for side effects and 

Build is failing b/c Travis tests need privileged creds to run. I rolled the commit over to new PR here: #751 please continue discussion over there.

CaseyFaist:
Closing in favor of #760 👍

PR #751, opened 2018-09-17
(Update): Adding .python-compile-env env var support for pip install

CaseyFaist:
Continued from #750 - our Travis tests need creds to run, so commit was rolled over from `tomasbielskis:master` so tests could run using my creds.

Recap: [Airflow](https://github.com/apache/incubator-airflow/blob/master/UPDATING.md#airflow-110) requires an env var. This is a bit tricky in an ephemeral build environment and (confirmed) can't be addressed by independently setting app env vars.

<3 Airflow

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -287,6 +287,7 @@ buildpack_sqlite3_install
 mtime "sqlite3.install.time" "${start}"
 
 # Source compile variables
+# shellcheck source=.python-compile-env
 if [ -f .python-compile-env ]; then
   source .python-compile-env
 fi
--- a/bin/compile
+++ b/bin/compile
@@ -286,11 +286,6 @@ source "$BIN_DIR/steps/sqlite3"
 buildpack_sqlite3_install
 mtime "sqlite3.install.time" "${start}"
 
-# Source compile variables
-# shellcheck source=.python-compile-env
-if [ -f .python-compile-env ]; then
-  source .python-compile-env
-fi
 
 # pip install
 # -----------
--- a/bin/compile
+++ b/bin/compile
@@ -286,6 +286,11 @@ source "$BIN_DIR/steps/sqlite3"
 buildpack_sqlite3_install
 mtime "sqlite3.install.time" "${start}"
 
+# Source compile variables
+if [ -f .python-compile-env ]; then
+  source .python-compile-env
+fi
+
 # pip install
 # -----------
 
--- a/bin/steps/pip-install
+++ b/bin/steps/pip-install
@@ -16,6 +16,15 @@ if [ ! "$SKIP_PIP_INSTALL" ]; then
 
     set +e
 
+    # Set SLUGIFY_USES_TEXT_UNIDECODE, for Airflow support
+    if [[ -r $ENV_DIR/SLUGIFY_USES_TEXT_UNIDECODE ]]; then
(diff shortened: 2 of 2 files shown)


bin/compile
```diff
@@ -286,6 +286,11 @@ source "$BIN_DIR/steps/sqlite3"
 buildpack_sqlite3_install
 mtime "sqlite3.install.time" "${start}"
 
+# Source compile variables
+if [ -f .python-compile-env ]; then
```
CaseyFaist: @tomasbielskis `-f` might be better as `-s`, so unless file exists and contains something, do nothing

tomasbielskis: Changed it, it's reflected in #750

tomasbielskis:
@CaseyFaist please let me know if there's anything else I could do to help move this forward!

theodoresiu:
Hi, would really like to push this PR forward so that we can update our version of Airflow on Heroku. We currently also have an opened ticket regarding this issue.

CaseyFaist:
@tomasbielskis that sounds about right. I'll pull down and test it with some airflow installs, will report back shortly

CaseyFaist:
for context, the errors seen with `heroku-toolbelt` are part of several CI and testing updates the buildpack needs; those updates are in process, but in order to get this feature unblocked I'm rolling the bare minimum needed into this PR 👍

theodoresiu:
I see that the same issue of missing file still exists. 😭All of the sourced files seem to exist in the bin directory. I wonder if moving the command to one of the step scrips would work or if something else needs to be done to the bash linter to allow it to pass. Otherwise the `-f` command seems correct. Thanks for all of your help @CaseyFaist and @tomasbielskis!

CaseyFaist:
hey @theodoresiu!

After rooting around, the issue is that shellchecking a file should only happen for files that live in this buildpack. The env file we're trying to add, by definition, is going to live in the user's repo - so that's why we're seeing that error.

I'm not giving up yet though! We may be able to use the [pre_compile](https://github.com/heroku/heroku-buildpack-python/blob/master/bin/steps/hooks/pre_compile) script as a path forward - this runs in a subenv so it doesn't help us pass env vars around as is, but a second build-env process might do the trick

CaseyFaist:
Hmm. Not the quick fix I was hoping it would be, so I'm going to pull out the travis updates and push those up separately. 

The Makefile is running the shellchecks, and shellcheck is supposedly active for everything in that `hook` step directory, so that shouldn't be the issue - my hunch is that we're not sourcing from the correct directory.

Heads up, I'm resetting this branch to your most recent commit to keep the branch clean 👍

CaseyFaist:
@tomasbielskis @theodoresiu looks like we have a winner, based off of https://github.com/heroku/heroku-buildpack-python/pull/639/files. It's the opposite of generic, and I've instead opted to add support specifically for airflow to the pip-install step. I set this PR as the buildpack on a test app with `apache-airflow==1.10` in the req file, and with SLUGIFY_USES_TEXT_UNIDECODE set in my Heroku configs, it successfully installed 🎉

theodoresiu:
Great, thanks a bunch Casey so will this be part of the standard python buildpack once merged?

CaseyFaist:
@theodoresiu it needs to go through the release process, so it won't be immediately accessible after merge. I'll need to move the revised changes to a clean PR as well, as review requested some changes. Release should be in production beginning of next week, Tuesday at the latest, as we have one release to push ahead of this one 👍

tomasbielskis:
@CaseyFaist Sounds great! Thank you for all your help on getting this through.

PR #756, opened 2018-10-03
Create pypy3-6.0.0

Marocco2:
This commit adds pypy3-6.0.0 to the buildpack


CaseyFaist:
Hi @Marocco2 - thanks for the PR! Unfortunately we don't currently have a support strategy for Python distributions other than the official branches. The runtime generation process is under review right now, too, so I'm hesitant to add support for a distribution prior to knowing what that update might look like.

I recommend adding an issue instead with the `feature request` tag - if there's significant community support for adding it, we'll consider formalizing a support strategy for the Pypy distributions of python 👍

PR #759, opened 2018-10-04
Update travis setup for Hatchet

CaseyFaist:
Hatchet has a Heroku CLI dependency that until recently, directly used the Toolbelt. This has changed, so the Travis test setup requires some changes before builds will succeed again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ dist: trusty
 rvm:
 - 2.4.4
 before_script:
+  - bash etc/ci-setup.sh
   - gem install bundler -v 1.16.2
   - bundle exec rake hatchet:setup_travis
 addons:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,16 +1,13 @@
...
   - gem install bundler -v 1.16.2
   - bundle exec rake hatchet:setup_travis
-addons:
-  apt:
-    sources:
-    - heroku
-    packages:
-    - heroku-toolbelt
+before_install:
+  - sudo bash etc/ci-setup.sh
```
jkutner: I know i keep my scripts in `etc/` but it's really the wrong place strictly speaking (`etc/` is for config files as i understand it). in theory, they should go in `bin/` the way the [Go buildpack does it](https://github.com/heroku/heroku-buildpack-go/tree/master/bin). I never liked cluttering `bin/` though, and I like keeping only buildpack API scripts in there. A better place for this script is maybe `opt/`. idk. i've never cared enough to move mine. So i'll leave it to you.

CaseyFaist: This is good to know 👍 I'll refactor to that if/when I convert to Heroku CI - the wrong place is good enough for now to get some other PRs unblocked.

jkutner: 👍

PR #760, opened 2018-10-05
Add Apache Airflow 1.10 Support

CaseyFaist:
This PR adds a check for SLUGIFY_USES_TEXT_UNIDECODE to the pip-install and pipenv install steps of the buildpack. Apache Airflow 1.10 needs this var set to "yes" to successfully install, but the pip-install steps don't natively have access to application env vars during the build.

This resolves #750


codingjoe:
That being said, shouldn't reverting this commit do the trick too?
https://github.com/heroku/heroku-buildpack-python/commit/920fefce5710e56b9908d2f819f585b83f5b2c0e

The commit message doesn't say much sadly, but I don't see when the environment should not be available during pkg installation. It is needed not only for airflow. Some C binding that call the compile during source package installation take environment variables to be configured. With that in mind, I would highly suggest to have the whole environment exposed in pip/pipenv. Unless of course there was a security concern connected to the commit mentioned earlier.

CaseyFaist:
Hey @codingjoe - yep, the prior art this PR is based on had no documentation I could find, and I found a few other behaviors that need documenting while working on this issue. Docs in the works 👍 

I agree, build steps having access to more details of the deploy environment is the more attractive long-term solution. The way the buildpack currently works though, the changes needed are not as self-contained as this PR, and a few other large projects are prioritized ahead of making env vars visible to the build process. Using a subenv did not resolve the issue unfortunately because of the scope problems that come with subprocesses in bash - same reason that a pre_compile script doesn't address the issues you mentioned. 

This PR unblocks users experiencing the bug reported in #750, and with minimal side effect to non-Airflow users; I'm happy with keeping this change self-contained for now.

codingjoe:
> Using a subenv did not resolve the issue unfortunately because of the scope problems that come with subprocesses in bash

@CaseyFaist how so, isn't that what the `sub_env` util does. I am really confused. Just for reference we are both taking about this, right? https://github.com/heroku/buildpack-stdlib/blob/472eff808d8a96d51cf1e27ada05dc8bbc65ba91/stdlib.sh#L100-L130

`pre_compile` doesn't expose the environment, because it is sourced as well and does not use the `sub_env` util.

> the prior art this PR is based on had no documentation I could find

uhm


@jkutner thanks, maybe it does make sense to whitelist some environment variables. It certainly makes sense for the geo libraries, because they require only a single environment variable.

PR #761, opened 2018-10-05
Replaced hardcoded /app with $BUILD_DIR

duanehutchins:
This PR is to replace many "/app" references with $BUILD_DIR

Heroku builds the environment within $BUILD_DIR before copying it to /app upon success.

This is to address an issue with the python buildpack where it's building everything in the /app directory which causes issues with other buildpacks and exported environment variables which are correctly using the $BUILD_DIR directory.

All build files are now installed within $BUILD_DIR, and exported environment variables (in /export) are now pointing to folders under $BUILD_DIR.

The "/app/..." symlinks were left but may not be necessary for many builds.

Environment variables exported with .profile.d were not changed because these are post-build.  In the future, it may be worth changing these from /app to $HOME.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -87,6 +87,7 @@ export PATH=$PATH:$ROOT_DIR/vendor/:$ROOT_DIR/vendor/pip-pop
 unset GIT_DIR PYTHONHOME PYTHONPATH
 unset RECEIVE_DATA RUN_KEY BUILD_INFO DEPLOY LOG_TOKEN
 unset CYTOKINE_LOG_FILE GEM_PATH
+export PYTHONPATH="$BUILD_DIR"
 
 # Import the utils script, which contains helper functions used throughout the buildpack.
 # shellcheck source=bin/utils
@@ -101,7 +102,7 @@ source "$BIN_DIR/warnings"
 # to `/app`.
 # Symlinks are required, since Python is not a portable installation.
 # More on this topic later.
-mkdir -p /app/.heroku
+mkdir -p "BUILD_DIR/.heroku"
 
 # This buildpack programatically generates (or simply copies) a number of files for
 # buildpack machinery: an export script, and a number of `.profile.d` scripts. This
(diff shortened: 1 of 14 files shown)


duanehutchins:
FYI, This passed Travis-CI in my own repo's test (same commit): https://travis-ci.com/duanehutchins/heroku-buildpack-python/builds/87232752

I can't pass my own heroku api variables in Pull Requests though.

CaseyFaist:
Hi Duane! I'm circling around to evaluate this; can you resolve the merge conflicts and provide some specific reproduce steps so I can test?

CaseyFaist:
Circling back to this - going to resolve conflicts and run with privileged builds (mine). Thank you for the contribution (and the patience) @duanehutchins ❤️ 🌮 🎉

CaseyFaist:
Superceded by #868  🎉 ❤️ thanks for your contributions!

dzuelke:
The hard-coding of `/app` was intentional.

If Python binary modules are built during a `pip install`, any dynamic library references are baked into the resulting code, with their paths, and that would now be `/tmp/…`, which does not exist at runtime.

That's why a symlink is created between `/app/.heroku/python` and `/tmp/…/.heroku/python`: so that the build happens using the same directory structure that's used at runtime.

If any environment variable issues or incompatibilities with other buildpacks exist because of this, those should be addressed explicitly.

Generally, the pattern is for a buildpack to create its `/app/.heroku/WHATEVER` symlink, which is left in place at the end of the buildpack run, and then correctly `export` anything relevant (such as `$PATH`) for subsequent buildpacks (the PHP buildpack does the same thing).

dzuelke:
@duanehutchins can you please provide a more detailed rationale for this PR?

You wrote:

> This is to address an issue with the python buildpack where it's building everything in the /app directory which causes issues with other buildpacks and exported environment variables which are correctly using the $BUILD_DIR directory.

Can you please provide a reproduce case for that?

duanehutchins:
@dzuelke 

Sure and thank you.  It comes back to what you said in your comment a few hours ago:

> That's why a symlink is created between /app/.heroku/python and /tmp/…/.heroku/python:

When those symlinks are created from `$BUILD_DIR/.heroku/python` and `$BUILD_DIR/.heroku/vendor` to their respective `/app` counterparts, it overwrites everything that already exists in those folders.

So, if another buildpack uses either of those directories before this happens, they will be lost.

When I wrote this PR last year, I was having an issue where gdal library installed by cyberdelia/heroku-geo-buildpack (into `$BUILD_DIR/.heroku/vendor/`) was being erased by this python buildpack, and I traced it back to this cause. I also couldn't move the geo-buildpack to install after python because then the python would fail to install dependencies which required gdal. I also couldn't use the built-in gdal because the binary wasn't updated to include the new (at the time) Heroku-18 stack.

If you also notice, heroku released its own heroku/heroku-buildpack-geo later in 2018 which addressed this very problem by using the `$BUILD_DIR/.heroku-geo-buildpack/vendor/` path instead of `$BUILD_DIR/.heroku/vendor/`

This was some time ago, but there was also another issue involving the environment variables, but I can't remember it exactly at this point.

In reality, buildpacks shouldn't have to worry about other buildpacks erasing entire directories like this one does.

Issue #762, opened 2018-10-06
pip is upgraded to latest version when installing pipenv


jxltom:
https://github.com/heroku/heroku-buildpack-python/blob/d18f1fedd87c209d47969e5f22bf1d6767fcbb00/bin/steps/pipenv#L47

The pip is upgraded to latest version when installing pipenv. The pip version should be pinned to have stable and repeatable build environment.

Othewise, unexpected failures may happen even there is no update in codebase, such as in https://github.com/pypa/pipenv/issues/2924.

techalchemy:
If the buildpack is upgrading pip when installing pipenv, it is likely a consequence of using the default 'upgrade strategy', which before pip 10 was set to `eager`, but can be set to `only-if-needed` (which I see you are accurately capturing in your PR)

Technically, this is a bug on our end due to a gap in one of the libraries we broke out of pipenv in the last few months -- it was failing to respect our vendored / patched pip version.  That should be fixed in the new release

CaseyFaist:
After digging into this issue separately, I came to this same conclusion - 🌮 's for you all, thanks for your commits. I wish I could merge #763 directly, but unfortunately I have quite a bit of work to do still to make the tests more contributor-friendly. 

The solution can be found at #769; it's being prepped for review and I'm targeting a release this week 👍

techalchemy:
definitely hold off on upgrading pipenv until the next iteration so we can make sure the bugs are sorted out -- we had a lot of changes in the release so just want to make sure we capture everything major, but this release will bring a _lot_ of consistency to the resolver

Thanks for your patience and sorry for the issues! I am a consumer of heroku enterprise at work myself so I have some stake in making sure this works

PR #763, opened 2018-10-06
Do not upgrade system pip

jxltom:
Try to fix https://github.com/heroku/heroku-buildpack-python/issues/762

Pip is upgraded to lastet version during installing pipenv. This may cause unexpected circumstances in building process. This also makes later steps which relys on specific pip version fails.

### DIFF
--- a/bin/steps/pipenv
+++ b/bin/steps/pipenv
@@ -44,7 +44,7 @@ if [ ! "$SKIP_PIPENV_INSTALL" ]; then
         export PIPENV_VERSION="2018.5.18"
 
         # Install pipenv.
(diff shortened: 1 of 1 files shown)


CaseyFaist:
Hey @jxltom ! It took some digging as to how pipenv was causing the unexpected behavior seen on the platform, but this is the solution I'm moving forward with to resolve. Thank you for the PR 🌮

jxltom:
Great. Glad it helps!

CaseyFaist:
@jxltom Excellent context 👍 I'll ensure the fix conforms to that

Issue #764, opened 2018-10-08
python buildpack erases other vendor files while "clearing cache"


duanehutchins:
If the stack or SQLite3 changes, this buildpack removes all existing files in `.heroku/vendor` along with other several other directories: https://github.com/heroku/heroku-buildpack-python/blob/d3180d3245d08e2abe9f7bb265e9fd633ea29ae5/bin/steps/python#L49

This breaks installations from other buildpacks which installed into `.heroku/vendor` or another of those removed directories (e.g. https://github.com/cyberdelia/heroku-geo-buildpack/).

These files are erased by this buildpack in order to clear out the cache which itself copied into them during bin/compile: https://github.com/heroku/heroku-buildpack-python/blob/d3180d3245d08e2abe9f7bb265e9fd633ea29ae5/bin/compile#L166

I suggest that instead of copying and then removing the invalid cache files, this buildpack only copies the cache files if cache is valid.

pythdasch:
Hi I'm facing a similar problem, I'm using your Buildpack alongs  https://github.com/cyberdelia/heroku-geo-buildpack

and since 1 week, I cannot build anymore because it says that it doesn't find what installed geo-buildpack.

django.core.exceptions.ImproperlyConfigured: Could not find the GDAL library (tried "gdal", "GDAL", "gdal2.2.0", "gdal2.1.0", "gdal2.0.0", "gdal1.11.0", "gdal1.10.0", "gdal1.9.0"). Is GDAL installed? If it is, try setting GDAL_LIBRARY_PATH in your settings.

duanehutchins:
@pythdasch Set BUILD_WITH_GEO_LIBRARIES=1, and remove the geo buildpack, and this python buildpack should install those geo libraries for you.

Alternatively, you're welcome to switch your python buildpack to my own github fork, until this repo fixes the bug.

CaseyFaist:
This is not related to the vendoring issue, but for using GEO libraries, recommended way is to use this buildpack: https://github.com/heroku/heroku-geo-buildpack

On to vendoring - I see this issue, this is something we've had a chat or two around whether this should be a buildpack standard (and not just something Python should do) - several other buildpacks handle caches this way, but it's not explicitly required here: https://devcenter.heroku.com/articles/buildpack-api#caching

Can't promise when we'll see movement on this, but am interested in addressing this concern.

edmorley:
[...]

PR #765, opened 2018-10-08
Validate stack and python version prior to copying cache into build

duanehutchins:
This is to address issue #764. The cache directories are only copied if the cache is a valid match, rather than always copying the directories and then deleting them if the cache is not a valid match.


CaseyFaist:
closing - issue is still open, but will want to confer with other buildpack owners and agree on a standard (as mentioned in issue) before proceeding in a specific direction.

PR #767, opened 2018-10-08
Dont upgrade pipenv

CaseyFaist:
Pipenv currently updates both pipenv and pip to most recent version on deploy as default behavior. To match pip behavior, this PR locks pipenv to version 2018.7.1, uses the buildpack default pip version, and allows for user override.

This solution is favored over #763, which would still update pipenv on deploy without user explicitly choosing that. Also inspired by #732

### DIFF
--- a/bin/steps/pipenv
+++ b/bin/steps/pipenv
@@ -9,14 +9,11 @@ set -e
 if [[ -f Pipfile.lock ]]; then
     if [[ -f .heroku/python/Pipfile.lock.sha256 ]]; then
         if [[ $(openssl dgst -sha256 Pipfile.lock) == $(cat .heroku/python/Pipfile.lock.sha256) ]]; then
-            # Measure that we're using Pipenv.
-            mcount "tool.pipenv"
 
             # Don't skip installation of there are git deps.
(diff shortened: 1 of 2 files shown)


bin/steps/pipenv
```diff
@@ -46,17 +41,24 @@ if [ ! "$SKIP_PIPENV_INSTALL" ]; then
             mcount "buildvar.PIP_EXTRA_INDEX_URL"
         fi
 
+        # PIP_EXTRA_INDEX_URL allows for an alternate pypi URL to be used.
+        if [[ -r $ENV_DIR/DEFAULT_PIPENV_VERSION ]]; then
```
sigmavirus24: I'm wondering if we want this to be quoted.

bin/steps/pipenv
```diff
@@ -46,17 +41,24 @@ if [ ! "$SKIP_PIPENV_INSTALL" ]; then
...
+            DEFAULT_PIPENV_VERSION="$(cat "$ENV_DIR/DEFAULT_PIPENV_VERSION")"
```
sigmavirus24: Does this allow for overriding the default version?

PR #768, opened 2018-10-09
Refactor env variable handling in build steps & add docs

codingjoe:
Ref #760 

### Manually tested

- [x] Airflow install
- [x] GEO libs without setting the pass explicitly in Django
- [x] extra package index

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,3 +61,20 @@ Runtime options include:
 - `python-3.7.0`
 - `python-3.6.6`
 - `python-2.7.15`
+
+Buildpack specific environment variables
+----------------------------------------
+
+- `DISABLE_COLLECTSTATIC` (default `0`):
+  Should the buildpack detects a Django app, it will automatically collect the static files for your.
+  If set to `1` it will deactivate this behavior.
+  
+  
+- `BUILD_WITH_GEO_LIBRARIES` (default `0`):
+  If set to `1` the buildpack will install GDAL, GEOS and PROJ.
+  
+- `PIP_EXTRA_INDEX_URL` (default ``):
+  Can be used to provide additional pip package repositories.
+  
+- `SLUGIFY_USES_TEXT_UNIDECODE` (default `unset`):
+  Required to install Airflow version 1.10 and later.
--- a/bin/compile
+++ b/bin/compile
@@ -269,7 +269,7 @@ source "$BIN_DIR/steps/pylibmc"
 # Support for Geo libraries. This is deprecated functionality, only functional on cedar-14.
 # It is undocumented.
 # shellcheck source=bin/steps/geo-libs
-sub_env "$BIN_DIR/steps/geo-libs"
+source "$BIN_DIR/steps/geo-libs"
 
 # GDAL support.
 # This is part of the Geo support.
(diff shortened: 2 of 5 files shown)


CaseyFaist:
Thank you for the contribution! I agree with the intent of the changes and appreciate the refactors, but won't be moving forward with this PR for a couple of reasons.

First, I'm currently planning some significant restructuring in the coming months. Since the proposed changes could likely have edge case effects I'd need to find, test for and document, it's not efficient to invest the time to vet these updates for potential side effects when the surrounding processes themselves could be overhauled somewhat soon.

For the documentation updates, keeping documentation in comments by the implementation is not ideal. Comments are important for context on the code decisions, but anyone looking for a user-friendly feature list will not have a good experience if features are documented only where they are implemented. Different options are being considered to solve this problem, but no decision as of yet.

codingjoe:
> Thank you for the contribution! I agree with the intent of the changes and appreciate the refactors, but won't be moving forward with this PR for a couple of reasons.

Sure, as you wish.

> First, I'm currently planning some significant restructuring in the coming months. Since the proposed changes could likely have edge case effects I'd need to find, test for and document, it's not efficient to invest the time to vet these updates for potential side effects when the surrounding processes themselves could be overhauled somewhat soon.

That sounds exciting I can't wait!

> For the documentation updates, keeping documentation in comments by the implementation is not ideal. Comments are important for context on the code decisions, but anyone looking for a user-friendly feature list will not have a good experience if features are documented only where they are implemented. Different options are being considered to solve this problem, but no decision as of yet.

I did add the documentation to the readme, as I obviously can't edit the articles in the Heroku knowledbge base. I do agree tho. Good point.

One last thing:
I would recommend to move ahead with 735663e tho. It is simply fixing a bug that has been effecting everyone using Django and PostGIS. See #398.

thanks joe

codingjoe:
I could open up another PR maybe also change they way I import the environment variable to how it is done in some other places? What do you think?

CaseyFaist:
Re: env vars - I'm planning on reviewing how the buildpack accesses environment variables, with this bit from the Node buildpack under consideration as a possible direction: https://github.com/heroku/heroku-buildpack-nodejs/blob/7df64870c0cbc1d08c3f83bd5060c5d522126729/lib/environment.sh#L52

In the short term, my focus is going to be on expanding test coverage - but if you could pull out the bugfix and wouldn't mind updating the environment variable import to match other env var loads to make the planned refactor easier, that'd definitely be worth investigating for me. Thanks for responding! I appreciate your patience while I get up to speed with the issue you linked 👍 👍 👍

codingjoe:
> In the short term, my focus is going to be on expanding test coverage - but if you could pull out the bugfix and wouldn't mind updating the environment variable import to match other env var loads to make the planned refactor easier, that'd definitely be worth investigating for me. Thanks for responding! I appreciate your patience while I get up to speed with the issue you linked 👍 👍 👍

My pleasure, done #770 

More tests would certainly be great. Might I make a bold suggestions? Why are all buildpacks written in Bash not Python? Python is installed on [all images anyways](https://devcenter.heroku.com/articles/stack-packages) since Ubuntu switched may launch scripts from Bash to Python a long time ago. Maybe this is a path worth setting into? Testing python is certainly more rewarding than bash, don't you agree?

PR #770, opened 2018-10-15
Ref #398 -- Expose GDAL_LIBRARY_PATH & GEOS_LIBRARY_PATH to build

codingjoe:
`steps/geo-libs` was executed in a sub_env and not sourced. The
trailing export statement did have no effect and would not be
available for Django's `collectstatic` or any post_compile hook.


codingjoe:
I just tested it manually again, works fine for me.

codingjoe:
@CaseyFaist i just ran a test. This at least exposed environment variables during build, but it doesn't fully resolve the Django issue. For that a Python dependency is missing in the runtime image, see https://github.com/heroku/stack-images/issues/118

duanehutchins:
@codingjoe Here's a snippet I used in Django's `settings.py` that resolved this issue and also your heroku/stack-images#118 ticket:
```
# If we aren't building in /app, then fix the environment variables which assumed we were
BUILD_DIR = os.environ.get("BUILD_DIR", "/app")
if BUILD_DIR != "/app":
    lib_re = re.compile(r'(?:^|(?<=:))/app(?=/)', re.ASCII | re.MULTILINE)
    lib_vars = set(["LIBRARY_PATH","LD_LIBRARY_PATH","GEOS_LIBRARY_PATH","GDAL_LIBRARY_PATH","PROJ4_LIBRARY_PATH","GDAL_DATA","GDAL"])
    for lib_var in lib_vars:
        if lib_var in os.environ and os.environ.get(lib_var, None):
            os.environ[lib_var] = lib_re.sub(BUILD_DIR, os.environ.get(lib_var))

# Glob for archive lib name within path provided or else LD_LIBRARY_PATH
def lib_find_archive(lib_paths, lib_name):
    import subprocess, glob

    if lib_paths is not None:
        if not lib_paths or os.path.isfile(lib_paths):
            # Quit if path is empty or points to an actual file
            return lib_paths or None

        elif not os.path.isdir(lib_paths):
            # Separate archive path from name
            lib_name = os.path.basename(lib_paths) or lib_name
            lib_paths = os.path.dirname(lib_paths) or None

    # lib_name must be populated
    if not lib_name: return None

    # Glob search within lib_path or else LD_LIBRARY_PATH
    if lib_paths:
        lib_paths = [lib_paths]
    elif(os.environ.get("LD_LIBRARY_PATH", None)):
        lib_paths = os.environ.get("LD_LIBRARY_PATH", None).split(':')
    else:
        return None

    # Glob-seach each directory for lib_name
    for lib_path in lib_paths:
        if not (lib_path and os.path.isdir(lib_path)):
            continue

        # Find matching archive files
        lib_path_glob = os.path.join(lib_path, lib_name)
        lib_paths_globbed = glob.glob(lib_path_glob)

        # Use shortest-length match
        if lib_paths_globbed: return sorted(lib_paths_globbed, key=len)[0]

    # Archive not found
    return None

# Collect library paths and correct each, if necessary
GDAL_LIBRARY_PATH = lib_find_archive( os.environ.get('GDAL_LIBRARY_PATH', None), 'libgdal.so*' )
GEOS_LIBRARY_PATH = lib_find_archive( os.environ.get('GEOS_LIBRARY_PATH', None), 'libgeos_c.so*' )
PROJ4_LIBRARY_PATH = lib_find_archive( os.environ.get('PROJ4_LIBRARY_PATH', None), 'libproj.so*' )
```

The top part translates the paths to the build_dir during heroku builds. Depending on your setup, this might not be needed.

If GDAL_LIBRARY_PATH is set to an existing file, it's used as-is. If it's not set, then LD_LIBRARY_PATH is searched for the libgdal.so* archive. The same logic is then used to find the GEOS and PROJ4 archives. 

IMO, Django should use a search like this if `ld` is not installed, rather than just giving up as it does now.

Good Luck.

codingjoe:
@duanehutchins thanks for the reply. We did work around the issue by setting the `GDAL_LIBRARY_PATH` setting in Django directly. Should Django handle this issue better, sure. But still I believe `ld` should be available in the stack images for proper Python support. This patch is also needed, to have the buildpack at least work as expected and set the environment correctly as expected both during run and build time.

dzuelke:
@codingjoe `ld` is available on the heroku-18 stack!

dzuelke:
As a matter of fact, it's on heroku-16 as well!?

codingjoe:
@dzuelke it was added 2 month ago. I missed that because the issue was not referenced in the commit. Anyhow, that means it should work tho. I will retest it again asap.

codingjoe:
Yes, that seems to work now!

PR #771, opened 2018-10-22
feat: add ability to override where lang-common gets pulled from

josegonzalez:
This follows the pattern set in place for `VENDOR_URL`, allowing folks to override where the common stdlib.sh gets pulled from.

### DIFF
--- a/bin/utils
+++ b/bin/utils
@@ -2,9 +2,19 @@
 shopt -s extglob
 shopt -s nullglob
 
+# Set the Common lib's internet target for downloading common dependencies.
+# The user can provide BUILDPACK_COMMON_URL to specify a custom target.
+# Note: this is designed for non-Heroku use, as it does not use the user-provided
+# environment variable mechanism (the ENV_DIR).
+COMMON_VENDOR_URL="https://lang-common.s3.amazonaws.com"
+if [[ -n ${BUILDPACK_COMMON_URL:-} ]]; then
+    COMMON_VENDOR_URL="$BUILDPACK_COMMON_URL"
+fi
+export COMMON_VENDOR_URL
+
 # The standard library.
 if [[ ! -f  /tmp/stdlib.sh ]]; then
-  curl --retry 3 -s https://lang-common.s3.amazonaws.com/buildpack-stdlib/v8/stdlib.sh > /tmp/stdlib.sh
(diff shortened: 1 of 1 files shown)


josegonzalez:
@CaseyFaist would it be possible to also get this pulled in? For... various reasons, one of my freelance clients wants to centralize their CI on GCP - including network calls for reasons I couldn't explain - and they have a few apps on Heroku which would benefit from this change :)

jmorrell:
I'm hesitant to allow user config to modify where dependency code comes from... I'm not certain how we can provide any guarantees about not breaking users who do this.

I'm not sure I understand the use case here. Do the clients just want to avoid a call to S3?

CaseyFaist:
Yeah @josegonzalez , I need more context than that to feel confident about allowing a core dependency of the buildpack to be replaced. The line `"${COMMON_VENDOR_URL}/buildpack-stdlib/v8/stdlib.sh" > /tmp/stdlib.sh` to me sounds like they've forked the standard library internally and intend to stay at version 8, which would mean I couldn't refactor this to update to latest version without causing breaking side effects for them.

josegonzalez:
So for whatever reason they are going through an audit, and part of that audit means that the artifact that was tested needs to be what they actually ship. The auditor in this case is pretty ornery about where the CI testing runs - it was a pain to get them to even agree to allowing the apps to continue living on Heroku and letting us run CI on our own, and I kinda don't want to have that conversation anymore.

It's more or less replicating what _should_ be in Heroku CI, but within their own, existing CI server. Thus, they need to control the artifacts that get installed so that they can 100% know that the application that is deployed at any given time is the same thing they had at a previous time. Network calls to services/resources outside of their control are more or less forbidden, hence why this needs to be vendored.

It's annoying that I need to essentially ship a buildpack docker image to heroku - I'm very much "what the actual??!?!" on this too - but at least I get to use the rest of Heroku for the apps, so anything I can do to reduce the network calls is appreciated.

> Semi-related, I am having a fun time figuring out how to essentially vendor the pip and setuptools installations to point to a custom GCS bucket, but I'm taking baby steps here :)

I hope that answers your questions!

CaseyFaist:
Thank you for the context @josegonzalez ! Unfortunately, I don't think we will be moving forward with this functionality as the scope seems pretty specific and not very applicable to the larger audience of this buildpack. This use case seems to be out of scope of the buildpack's intended functionality, but I encourage you to fork and extend as you need for these clients.

josegonzalez:
Sounds good, thanks for the update!

PR #772, opened 2018-10-22
V 145

CaseyFaist:
Python 3.7.1 and 3.6.7 dropped over the weekend!

Also rolling 3.5.6 in here.

In favor of #749


josegonzalez:
@CaseyFaist would be great if we could also include https://github.com/heroku/heroku-buildpack-python/pull/774 as it should handle the comment [here](https://github.com/heroku/heroku-buildpack-python/pull/749#pullrequestreview-153093812).

CaseyFaist:
@josegonzalez good catch, I'll pull #774 in 👍

CaseyFaist:
@schneems some tests are expected to fail as these versions are not available in prod yet, but that looks like the version messaging is too strict, so is related in this case. Addressing

CaseyFaist:
K. Cedar 14 stack failure for tests is unrelated, all others pass 👍 this is ready for review and release ❤️

bin/steps/python
```diff
@@ -7,41 +7,49 @@ PYTHON_VERSION=$(cat runtime.txt)
...
 VENDORED_PYTHON="${VENDOR_URL}/runtimes/$PYTHON_VERSION.tar.gz"
 
-if [[ $PYTHON_VERSION =~ ^python-2 ]]; then
-  if [[ "$PYTHON_VERSION" != "$LATEST_2" ]]; then
-    puts-warn "The latest version of Python 2 is $LATEST_2 (you are using $PYTHON_VERSION, which is unsupported)."
-    puts-warn "We recommend upgrading by specifying the latest version ($LATEST_2)."
-    echo "       Learn More: https://devcenter.heroku.com/articles/python-runtimes"
-  else
-    echo "       Using supported version of Python 2 ($PYTHON_VERSION)"
+SECURITY_UPDATE="Python has released a security update! Please consider upgrading to"
```
jmorrell: Are there tests around these security warnings?

bin/steps/python
```diff
@@ -7,41 +7,49 @@ PYTHON_VERSION=$(cat runtime.txt)
...
+      echo "       Learn More: https://devcenter.heroku.com/articles/python-runtimes"
+    fi
   fi
-else
-  if [[ $PYTHON_VERSION =~ ^python-3 ]]; then
-    if [[ $PYTHON_VERSION =~ ^python-3.7 ]]; then
-      if [[ "$PYTHON_VERSION" != "$LATEST_37" ]]; then
-      puts-warn "The latest version of Python 3.7 is $LATEST_37 (you are using $PYTHON_VERSION, which is unsupported)."
-      puts-warn "We recommend upgrading by specifying the latest version ($LATEST_37)."
+  if [[ "$PYTHON_VERSION" == $PY36* ]]; then
```
jmorrell: Semver matching functionality in the stdlib would be a definite improvement here

CaseyFaist: agreed! I'd love to extract all of this to stdlib to dry up the buildpack

PR #774, opened 2018-10-31
fix: drop broken version check

josegonzalez:
Due to how the version checks work - via sorting, not actual comparison - this previously resulted in _always_ installing sqlite3, even though it was already bundled for lower versions of python. The second version check also encompasses 3.7.0+, so there is no need to respecify the check.


CaseyFaist:
This is rolled into #772 👍 🌮 ❤️

Issue #740, opened 2018-08-09
Support pip10


CaseyFaist:
We are currently on 9.0.2 - let's update

cjolowicz:
Yes, please. This should make it possible to install projects using [poetry](https://poetry.eustace.io), or any other build backend, without requiring other changes to the buildpack:

```
pip install .
```

From the [pip reference guide](https://pip.pypa.io/en/stable/reference/pip/#pep-517-and-518-support):

> As of version 10.0, pip supports projects declaring dependencies that are required at install time using a pyproject.toml file, in the form described in PEP 518. When building a project, pip will install the required dependencies locally, and make them available to the build process. Furthermore, from version 19.0 onwards, pip supports projects specifying the build backend they use in pyproject.toml, in the form described in PEP 517.

Issue #776, opened 2018-11-05
[Feature reqeust] Let python buildpack can access private repo with custom environment variables


bruce3557:
In production build, we often need to hide some credentials (oauth token, username / password ...) to environment variables and use them in building resources like `setup.py` or `requirements.txt`.

After trying the code like this
```python
GITHUB_USER = os.getenv("GITHUB_USER")
GITHUB_PASSWD = os.getenv("GITHUB_PASSWD")
...
install_requires=[
    ...
    "https://{github_user}:{github_password}@github.com/codementordev/cm-data-utils.git".format(
        github_user=GITHUB_USER,
        github_password=GITHUB_PASSWD
    )
    ...
]
```
and it doesn't work.

Is it possible to support this kind of features in the official buildpack? (or something like this)

kennethreitz:
environment variables are not exposed during installation. This is for security purposes, as setup.py's can theoretically do anything, including sending your credentials off to some server somewhere.

kennethreitz:
you can just commit a revocable token to your source tree

edmorley:
Another option is:
https://github.com/timshadel/heroku-buildpack-github-netrc

bruce3557:
@kennethreitz @edmorley  Thanks for the answers!

Agree that the `setup.py` has some security issues.

In fact, we also discussed above solutions and have security concerns on them:
- In general, we avoid to commit credentials-like stuffs in code.
- The things why we don't use third-party buildpacks is about the maintenance and security.

Currently we solve the issue by generating private repo links before executing `pip install` but still want to discuss how can make this kind of process more efficient.

edmorley:
> The things why we don't use third-party buildpacks is about the maintenance and security.

For the above buildpack, one of the forks is hosted within the Heroku GitHub org, if that helps:
https://github.com/heroku/heroku-buildpack-github-netrc
(plus with any buildpack you can always fork to your own org and point the app at the forked repo)

bruce3557:
Thanks for your sharing! It is helpful!
We traced the code and have another solution on it.
We use `pre_compile` file to generate private repo link with credentials and it works~

CaseyFaist:
[...]

PR #791, opened 2018-12-06
Update to latest pip

josephlbarnett:
Also allow GIT_OAUTH_KEY environment variable to be passed through to pip.

-----------------------
We wanted to be able to remove github oauth key values from our requirements.txt files that are checked into source control.  With later versions of pip (10+), it can read them from an environment variables.  This PR updates to the current latest pip (18.1), fixing https://github.com/heroku/heroku-buildpack-python/issues/740, and allows passthrough of a GIT_OAUTH_KEY env variable from heroku in order to meet this use case.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -66,7 +66,7 @@ PY27="python-2.7"
 # Which stack is used (for binary downloading), if none is provided (e.g. outside of Heroku)?
 DEFAULT_PYTHON_STACK="cedar-14"
 # If pip doesn't match this version (the version we install), run the installer.
-PIP_UPDATE="9.0.2"
+PIP_UPDATE="18.1"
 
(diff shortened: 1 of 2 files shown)


CaseyFaist:
Hi @josephlbarnett ! Thanks for the PR. I have some context and questions for you:

- Currently, pip version 9 is set for both pip install and pipenv install flows. The Pipenv version currently locked in the buildpack does not work with modern pip (this was surfaced in a buildpack bug back in October). Updating Pip is necessary, but this update as-is would require significant communication with current pipenv users to avoid breaking changes. 

- Continuing to add sourced env vars is not the ideal pattern moving forward for the buildpack, although it has been necessary for blocking issues. Did you explore the private auth flow discussed in #776 ? What worked for you and what didn't?

josephlbarnett:
Thanks for the reply Casey, we hadn't looked at #776, will do so now.

PR #793, opened 2018-12-19
Update requirements.txt

YashviP:


CaseyFaist:
These dependencies are not required for buildpack functionality, so closing

PR #795, opened 2018-12-24
Remove shellcheck travis install

jmorrell:
Shellcheck is now installed on travis by default so we don't need to install it manually anymore: https://github.com/koalaman/shellcheck#travis-ci


CaseyFaist:
This was cherrypicked into recent release 👍

PR #799, opened 2019-02-15
Build cleanup

dzuelke:
- [x] move `Dockerfiles` to `builds/` so it's all self-contained in there
- [x] `Dockerfiles` for all stacks and clean them up
- [x] Docker/build `README`
- [x] reuse logic in `runtimes/pypy-…` formulae
- [x] reuse logic in `runtimes/pypy3-…` formulae
- [x] reuse logic in `runtimes/python-…` formulae
- [x] cedar 14 reinstated for users still migrating

Moved to other work item: fix the `sqlite` mess


builds/heroku-18.Dockerfile
```diff
@@ -3,7 +3,8 @@ FROM heroku/heroku:18-build
 WORKDIR /app
 ENV WORKSPACE_DIR="/app/builds" \
     S3_BUCKET="lang-python" \
-    S3_PREFIX="heroku-18/"
+    S3_PREFIX="heroku-18/" \
+    STACK="heroku-18"
```
CaseyFaist: Can you explain what the stack var is for @dzuelke? is this so we don't have to specify it in the test steps like in the Makefile?

dzuelke: It's mostly so that it's available to formulae in case they have to do different things depending on the stack (say `apt-get install …-dev` packages)

CaseyFaist:
Updates: sqlite fix will be broken out into separate PR, and support for testing pre-release Python and other dependency binaries added.

dzuelke:
This is now missing all the cedar-14 specific stuff from earlier commits (README, Dockerfile)

CaseyFaist:
Unblocked by #839, now ready for final review

.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: ruby
-dist: trusty
+dist: bionic
```
jmorrell: Note: change this back

bin/compile
```diff
@@ -43,7 +43,6 @@ VENDOR_URL="https://lang-python.s3.amazonaws.com/$STACK"
 if [[ -n ${BUILDPACK_VENDOR_URL:-} ]]; then
     VENDOR_URL="$BUILDPACK_VENDOR_URL"
 fi
-export VENDOR_URL
```
jmorrell: Note: add this back

.travis.yml
CaseyFaist: 👍

bin/compile
CaseyFaist: 👍

.travis.yml
```diff
@@ -9,10 +9,12 @@ before_script:
   - bundle exec hatchet ci:setup
 
 script:
-  - docker build --pull --tag heroku-python-build-heroku-16 --file $(pwd)/Dockerfile .
-  - docker run --rm -e "STACK=heroku-16" heroku-python-build-heroku-16 bash $TESTFOLDER
-  - docker build --pull --tag heroku-python-build-heroku-18 --file $(pwd)/Dockerfile.heroku-18 .
-  - docker run --rm -e "STACK=heroku-18" heroku-python-build-heroku-18 bash $TESTFOLDER
+  - docker build --pull --tag travis-build-cedar-14 --file $(pwd)/builds/cedar-14.Dockerfile .
```
jmorrell: It's not an issue with this PR, but just documenting my concern that the tests on travis don't use the official images that the buildpack is deployed on

CaseyFaist: Noted. To be clear, the Dockerfiles *do* use the official images:
https://github.com/heroku/heroku-buildpack-python/blob/master/Dockerfile.heroku-18#L1

The dockerfiles don't pull empty ubuntu boxes and try to build from scratch or anything like that. In addition, the previous updates to the heroku-18 dockerfile files ensure that we don't deviate from the official image ( thanks again @dzuelke )

The issue is that the Python buildpack needs Python set up and the Buildpack in the environment, but using docker means we can't use the Python buildpack to install Python or set up the environment. It's an ouroboros problem. See the dockerfiles for more details on what setup is required to build binaries and run tests 👍

jmorrell: > the Dockerfiles do use the official images

That is a good point, but it's still a different environment than it's deployed on

> The issue is that the Python buildpack needs Python set up and the Buildpack in the environment, but using docker means we can't use the Python buildpack to install Python or set up the environment. It's an ouroboros problem.

Would using the `build` images fix this? If not, how does python get installed to run the builds in production currently?

PR #802, opened 2019-02-21
track usage of older feature steps

CaseyFaist:
This PR uses the `mcount` function provided by the standard library to count how often the older legacy steps are used at build time

### DIFF
--- a/bin/steps/geo-libs
+++ b/bin/steps/geo-libs
@@ -22,6 +22,7 @@ source "$BIN_DIR/utils"
 
 # If GDAL exists within requirements, use vendored gdal.
 if [[ "$BUILD_WITH_GEO_LIBRARIES" ]]; then
+  mcount "buildvar.BUILD_WITH_GEO_LIBRARIES"
 
   if [ ! -f ".heroku/vendor/bin/proj" ]; then
     echo "-----> Bootstrapping gdal, geos, proj."
@@ -31,9 +32,11 @@ if [[ "$BUILD_WITH_GEO_LIBRARIES" ]]; then
     curl "$VENDORED_GEOS" -s | tar zxv -C .heroku/vendor &> /dev/null
     curl "$VENDORED_PROJ" -s | tar zxv -C .heroku/vendor &> /dev/null
 
+    mcount "steps.vendor.geo_libs"
     # Copy libjasper from build image to slug.
     if [[ "$STACK" == "heroku-16" ]]; then
       cp /usr/lib/x86_64-linux-gnu/libjasper.so* ".heroku/vendor/lib/."
+      mcount "steps.vendor.libjasper"
     fi
 
   fi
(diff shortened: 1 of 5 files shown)


bin/steps/geo-libs
```diff
@@ -22,6 +22,7 @@ source "$BIN_DIR/utils"
 
 # If GDAL exists within requirements, use vendored gdal.
 if [[ "$BUILD_WITH_GEO_LIBRARIES" ]]; then
+  mcount "buildvar.BUILD_WITH_GEO_LIBRARIES"
```
jmorrell: Using all caps in `mcount` is odd. Did you mean `buildvar.$BUILD_WITH_GEO_LIBRARIES` maybe?

bin/steps/mercurial
```diff
@@ -4,6 +4,6 @@
 if [[ -f "requirements.txt" ]]; then
```
jmorrell: I am curious as to why mercurial was special-cased

bin/steps/nltk
```diff
@@ -27,6 +27,8 @@ if sp-grep -s nltk; then
         python -m nltk.downloader -d "$BUILD_DIR/.heroku/python/nltk_data" "${nltk_packages[@]}" | indent
         set_env NLTK_DATA "/app/.heroku/python/nltk_data"
 
+        mcount "buildvar.nltk_packages_definition"
+        mcount "steps.nltk"
```
jmorrell: Are these two going to always be the same?

jmorrell: You are using `steps.vendor.foo` in other cases, but `steps.nltk` here. Is that intentional?

bin/steps/mercurial
```diff
@@ -4,6 +4,6 @@
...
 	if (grep -Fiq "hg+" requirements.txt) then
 		/app/.heroku/python/bin/pip install  mercurial | cleanup | indent
+		mcount "steps.mercurial"
```
jmorrell: You are using `steps.vendor.foo` in other cases, but `steps.mercurial` here. Is that intentional?

bin/steps/nltk
jmorrell: Oh, they might be the same, but you can make different queries `buildvar.*` and `steps.*`

bin/steps/geo-libs
CaseyFaist: I started doing that for buildvars here: https://github.com/heroku/heroku-buildpack-python/blob/master/bin/steps/pip-install#L28

Is there a standard I can follow? I can update the other buildvars to be internally consistent with it if so

jmorrell: No standard, just jumped out in this diff :)

bin/steps/nltk
CaseyFaist: "vendor" is for anything we're intentionally vendoring. We don't build our own NLTK, just download it, and so it's just a step (like mercurial). 

This is so I can aggregate on multiple dimensions, such as get all `steps`, get all `vendored` steps, get all `buildvars`, or get exactly each one in comparison

CaseyFaist: I missed being consistent on this buildvar output tho - fixing

PR #805, opened 2019-03-04
HOTFIX for cedar-14 python builds

dzuelke:
This temporarily hardcodes the `libsqlite3` version check return value to a package version available at http://archive.ubuntu.com/ubuntu/pool/main/s/sqlite3/.

This may break again at any time for this stack or others.

Just a quick workaround for the next few days until we have the sqlite3 installation refactored to only use headers and link to them for `pip install` using the right env vars.


hone:
thanks @CaseyFaist !! <3

PR #809, opened 2019-03-11
Python release prep

CaseyFaist:
This is final release prep for 2.7.16!

This PR:

- Isolates the default python versions so `compile` and `tests` reflect updates with one change
- Updates the default Python 2.7 version to reflect the new binary (necessary for release)


bin/default_pythons
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+DEFAULT_PYTHON_VERSION="python-3.6.8"
```
jmorrell: You can `export` in-line here unless this is a shellcheck thing I'm unaware of

```
export DEFAULT_PYTHON_VERSION="python-3.6.8"
```

bin/default_pythons
```diff
@@ -0,0 +1,10 @@
...
+LATEST_36="python-3.6.8"
+LATEST_37="python-3.7.2"
+LATEST_35="python-3.5.6"
+LATEST_34="python-3.4.9"
+LATEST_27="python-2.7.16"
+
+export DEFAULT_PYTHON_VERSION LATEST_37 LATEST_36 LATEST_35 LATEST_34 LATEST_27
```
jmorrell: I don't think you actually need to export the env vars at all unless they are going to be read by a separate process. It's unlikely to hurt though.

bin/compile
```diff
@@ -68,8 +61,8 @@ DEFAULT_PYTHON_STACK="cedar-14"
 # If pip doesn't match this version (the version we install), run the installer.
 PIP_UPDATE="9.0.2"
 
-export DEFAULT_PYTHON_VERSION DEFAULT_PYTHON_STACK PIP_UPDATE
-export LATEST_27 LATEST_36 LATEST_37 LATEST_35 LATEST_34
+export DEFAULT_PYTHON_STACK PIP_UPDATE
+# export DEFAULT_PYTHON_VERSION LATEST_27 LATEST_36 LATEST_37 LATEST_35 LATEST_34
```
jmorrell: Did you mean to delete this line?

test/run
```diff
@@ -95,25 +99,21 @@ testPylibmc() {
...
 testPython2_warn() {
     compile "python2_warn"
-    if [[ $STACK = "heroku-18" ]]; then
-      assertCapturedError
-    else
-      assertCaptured "python-2.7.14"
-      assertCaptured "security update!"
-      assertCapturedSuccess
-    fi
+    assertCaptured "python-2.7.15"
```
jmorrell: Did you want to add a variable for this too? I suppose it should always warn on `2.7.15` 🤔

test/run
```diff
@@ -1,5 +1,9 @@
 #!/usr/bin/env bash
 
+# Default Python Versions
+# shellcheck source=bin/default_pythons
+source "$BIN_DIR/default_pythons"
```
jmorrell: At first I was 🤔 , but then I was 😃

bin/compile
CaseyFaist: whoops yep

test/run
CaseyFaist: yeah, I left it as it's now an old version

PR #814, opened 2019-04-03
Python 3.7.3 release

CaseyFaist:
A new Python binary is available on the Heroku stack! 

Plus some typo corrections in changelog and more test updates

### DIFF
--- a/test/run
+++ b/test/run
@@ -2,7 +2,7 @@
 
 # Default Python Versions
 # shellcheck source=bin/default_pythons
-source "$BIN_DIR/default_pythons"
+source "bin/default_pythons"
 
 testAirflow() {
   export SLUGIFY_USES_TEXT_UNIDECODE="yes"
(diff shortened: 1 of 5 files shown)


test/run
```diff
@@ -2,7 +2,7 @@
 
 # Default Python Versions
 # shellcheck source=bin/default_pythons
-source "$BIN_DIR/default_pythons"
+source "bin/default_pythons"
```
jkutner: was this on purpose?

CaseyFaist: It was, the file wasn't being found before the change. should I have changed the shellcheck instead?

jkutner: oh, so this just never worked before?

CaseyFaist: Yep, missed its effects in this test script due to false positives. I checked that it worked this time by running tests for all stacks with only one binary generated, it responds correctly to the failure case now

PR #816, opened 2019-04-09
Modularize package install tools and support poetry

bsdz:
This PR adds support for poetry and modularizes the package system with the potential ability to include other package managers in future.

It's not 100% perfect yet. Eg, could extract runtime.txt value from pyproject.toml file too and should perhaps clean the build cache if a different package manager introduced.


bsdz:
Looks like one of the automated checks tests whether a ruby environment works against a python buildpack. This is the only one that fails.

CaseyFaist:
Hi @bsdz! 

Thanks for the pr! ❤️ 🌮 This is a lot of changes at once, and I'm curious if we could focus in on just the method checks for pipenv and pip. There are several competing poetry support PRs open and some changes to the build processes going on, but I'm interested in the logic added to `utils` here.

I have to trigger tests from my account anyway, so if you have no objections, I can cherry-pick the relevant commits and move forward with this detection logic 👍

bsdz:
> I have to trigger tests from my account anyway, so if you have no objections, I can cherry-pick the relevant commits and move forward with this detection logic

Please do, you can take what you need. I look forward to seeing 1st class support for poetry (and perhaps other dependency tools too)

:-)

bsdz:
I recently switch from a buildpack based on this PR to using a helper buildpack instead:

https://github.com/moneymeets/python-poetry-buildpack.git

Issue #786, opened 2018-11-19
Bump pipenv version to 2018.11.14


macbeth322:
There was an issue in pipenv that was causing heroku deployments to fail because not all dependencies were getting installed sometimes. This was fixed in https://github.com/pypa/pipenv/pull/3070 and released in `pipenv==2018.11.14`. Is it possible to bump the version of pipenv that is used by the buildpack so we can get this fix?

techalchemy:
As a sidenote I will be releasing a bugfix release tonight and it may be advantageous to pin that one

FYI this seriously impacts my work builds -- I can't deploy to heroku due to pipenv pinning which is super unfortunate and has been a big annoyance for my team (and a good motivator to get things fixed)

jwin742:
Is there a reason this hasn't been done yet? Seems like it should be a very simple change now that the bug in pipenv has been fixed. Or at least make the version used configurable. 

It's getting kind of annoying for my team and I to be stuck on older versions of pipenv

CaseyFaist:
Hi there! some context:

Unfortunately, the current version of Pipenv would require users to uninstall and reinstall their local pipenv environments before being able to deploy again successfully. This is due to incompatibility with the lock files that is not autocorrected by pipenv's update function. Since this would require coordination between the release and all pipenv users, or would result in a temporarily opt-in upgrade (which, if anyone missed the communications around it, could lead to quite a lot of frustration), we're not looking at this update any time soon.

I am sorry about this frustration, however our version (2018.5.18) is also not as behind the current version (2018.11.26) as it has been since the most recent pipenv release. I'd like to see a bit more stability in the release cycle from the project before prioritizing this issue.

nikolaik:
> I am sorry about this frustration, however our version (2018.5.18) is also not as behind the current version (2018.11.26) as it has been since the most recent pipenv release. I'd like to see a bit more stability in the release cycle from the project before prioritizing this issue.

Hi! pipenv have since your comment received two releases and it's main maintainer has dedicated time set aside for pipenv work by it's employer. Hope to see you reconsider your stance on pipenv upgrades

edmorley:
@nikolaik Hi! Deciding the best way forwards for alternative package manager support (for example, should the buildpack support them natively, or via additional buildpacks; should we add poetry support, what to do with the currently out of date pipenv support etc) is high up my list. Pipenv now having more recent releases is something I will definitely factor into that decision :-)

PR #820, opened 2019-05-20
Upgrade Pipenv to 2018.11.26

lsmag:
Hi! This is related to https://github.com/heroku/heroku-buildpack-python/issues/786

We're using Pipenv at my team but we've had problems with the version this buildpack pinned, `2018.5.18`. In the end we leveraged the compile hooks to install a different version, because this was a blocker and I didn't know a better way to do it... aside from upgrading the version upstream, which is what this PR is about.

This upgrades Pipenv to `2018.11.26`, but also allows the variable to be overriden, similar to how it is done in [other sections of the code][1].

The only thing I'm still confused about is how to test the overriding part. I've read `test/run` a bit, and I think I can create a new `testOverridenPipenv` function that exports the variable, then unsets it after success. Will it be enough?

[1]: https://github.com/heroku/heroku-buildpack-python/blob/master/bin/compile#L20

### DIFF
--- a/bin/steps/pipenv
+++ b/bin/steps/pipenv
@@ -52,7 +52,7 @@ if [ ! "$SKIP_PIPENV_INSTALL" ]; then
             mcount "buildvar.SLUGIFY_USES_TEXT_UNIDECODE"
         fi
 
-        export PIPENV_VERSION="2018.5.18"
+        export PIPENV_VERSION="${PIPENV_VERSION:-2018.11.26}"
 
         # Install pipenv.
         # Due to weird old pip behavior and pipenv behavior, pipenv upgrades pip


lsmag:
Oops, the CI failed with something I'm not familiar with :S https://travis-ci.org/heroku/heroku-buildpack-python/jobs/534864670 

I'll need your help here to get this PR in a mergeable state, this has been blocking pipenv to be widely adopted at my company.

jerr0328:
Looks like PRs from non-members fails (secrets aren't added in, which makes sense because you don't want a malicious PR to expose those secrets in Travis). I do hope this can get merged, it would help me since I have the same problem as #704

lsmag:
If it helps merging faster, I can change it to `export PIPENV_VERSION="${PIPENV_VERSION:-2018.5.18}" `, so that current builds don't break and we're still allowed to customize Pipenv.

lsmag:
Huh, sorry to ask again, but what's the status on this? I'm willing to put more effort myself into fixing tests and whatever else is needed, but I will need some guidance here.

This is fundamental to allow the company I work for to use Pipenv across all teams.

CaseyFaist:
Hi there! some context:

Unfortunately, the current version of Pipenv would require users to uninstall and reinstall their local pipenv environments before being able to deploy again successfully. This is due to incompatibility with the lock files that is not autocorrected by pipenv's update function. Since this would require coordination between the release and all pipenv users, or would result in a temporarily opt-in upgrade (which, if anyone missed the communications around it, could lead to quite a lot of frustration), we're not looking at this update any time soon.

I am sorry about this frustration, however our version (2018.5.18) is also not as behind the current version (2018.11.26) as it has been since the most recent pipenv release. I'd like to see a bit more stability in the release cycle from the project before prioritizing this issue.

PR #823, opened 2019-06-06
Add SKIP_PIPENV to allow skipping Pipenv only (not pip)

PiDelport:
For some deployments it's desirable to have a `Pipfile` for development, but a conventional `requirements.txt` for deployment, without using Pipenv's `Pipfile` and `Pipfile.lock`.

There is an existing `SKIP_PIPENV_INSTALL` environment variable, but setting this skips both Pipenv and pip based installation.

This change adds an environment variable `SKIP_PIPENV` that skips only Pipenv, while still processing `requirements.txt` with pip normally.


CaseyFaist:
Hi @PiDelport !

Thank you for the pr! 🌮 ❤️ I appreciate your contributions. I'm going to punt on this for now, because I'm interested in solving this problem by providing a mechanism to specify which package manager tooling the buildpack should use (probably in runtime.txt) but please do open an issue so I can ensure this use case is represented in the tests for that feature

vityas:
@CaseyFaist it's been a year, and while you blocked the PR, you seem to have no progress on the issue.
Could we reopen the PR?

vityas:
For those who stumble on this thread in search of solution there is a workaround that might be useful to you

## possible workaround

you can set your `.slugignore` file (in the root) to ignore Pipfile, that way deployment will look for requirement txt, which means you have to have one.
```
# ignore pipenv
Pipfile
Pipfile.lock
```

PR #825, opened 2019-06-20
Remove sqlite

CaseyFaist:
Opt-in fix to SQLITE3 build errors on Heroku 16.

To opt in and use a Python version that was not built against SQLITE3, add the following config var to your application:

```
PYTHON_NO_SQLITE3=true
```

To test this on your app as a PR:
```
heroku buildpacks:set https://github.com/heroku/heroku-buildpack-python.git#remove-sqlite
```

Will Support:
Heroku 18
✅ Python 3.7.3
✅ Python 3.6.8
✅ Python 2.7.16

Heroku 16
✅ Python 3.7.3
✅ Python 3.6.8
✅ Python 2.7.16

Previously we explored the option to alter the sqlite dependency and vendoring flow - solution was workable but added maintenance burden. In addition, using SQLITE3 on Heroku is not advised due to the platform's ephemeral nature. For maintainability and to avoid breakage around an ideally unused system package that is not available in the base Heroku image on any current stack, I'm killin it ded.

Those who still wish to SQLITE3 despite being advised against using SQLITE3 on the platform may investigate using the [Apt buildpack](https://github.com/heroku/heroku-buildpack-apt), which is maintained by volunteers and does not come with official support from Heroku (but should work just fine).


allen-munsch:
When this gets merged would the `app.json` then look something like this ?:

```
"buildpacks": [
    { "url": "https://github.com/heroku/heroku-buildpack-python" }
],
 "env": {
     "PYTHON_NO_SQLITE3":  { "value": "true"}
}
```

Thanks

econner:
Possibly dumb question as I don't really understand the details here, but why isn't the default Python built without Sqlite3 if it is not recommended to use Sqlite3? Seems like users of sqlite3 would need to use a different buildpack anyway (apt as you suggest)?

To clarify,
I am running Python 2.7.16 on Heroku 16 and I am not using SQLite3 nor do I care about having support for it. To get my builds working should I use PYTHON_NO_SQLITE3 true or false?

Fwiw, I managed to get our builds working by forking the buildpack and commenting out the bit that [installs sqlite3](https://github.com/econner/heroku-buildpack-python/commit/cbf78e29ade9866a427386a2d412961f89f7b631) which it looks like this PR removes. When I go back to using the official buildpack though should I be using PYTHON_NO_SQLITE3 or not?

test/run
```diff
@@ -237,6 +238,14 @@ testGitEgg() {
   assertCapturedSuccess
 }
 
+testNoSqlite() {
+  export PYTHON_NO_SQLITE3="true"
```
sigmavirus24: Looking at the nodejs buildpack, I think this needs to look more like

```suggestion
echo "true" > $env_dir/NODE_VERBOSE
```

javaguy:
@CaseyFaist can you update on when this will get merged?

Also I echo @econner's feedback. the heroku buildpack is currently broken due to sqlite3 dependency and it will remain broken unless this option is added to our app.json or build environment, then why not make no sqlite3 dependency the default instead of requiring users to to opt-in?

kenneth-reitz:
Historically, at least, this will break all iPython usage on the platform (Django’s premium shell offering).

kenneth-reitz:
That is the primary reason for supporting it in the first place.

CaseyFaist:
going with #858 instead - closing as a result

PR #828, opened 2019-08-02
updating build pack for libxml/xmlsec1 artifact

mgan59:
For those that need a compiled SSO python-buildpack 😉 

* added more paths
* adding xmlsec1 artifact


mgan59:
my bad didn't realize this pr went against upstream, was meant for my local fork only.

PR #829, opened 2019-08-15
feat: add docs and make target for heroku-18 bob builds

josegonzalez:
This pull request adds a make target for building python packages under heroku-18. The dockerfile already existed, so a make target - plus the requisite docs - were added to simplify the process for users.


CaseyFaist:
Confirmed that tests pass locally, merging and releasing 🎉

PR #831, opened 2019-08-21
Remove cedar tests

CaseyFaist:
Cedar 14 is deprecated, and thus Travis tests and local docker tests failed on run.

Remove tests for deprecated stack to prevent false negatives:

- Remove Travis stack test for Cedar 14
- Remove make command test-cedar-14

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,10 +19,6 @@ jobs:
     services: docker
     env: STACK=heroku-16
     script: "./tests.sh"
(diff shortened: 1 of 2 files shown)


jmorrell:
There are a couple of more places where `cedar` is referenced if you want to do a deeper GC: https://github.com/heroku/heroku-buildpack-python/search?q=cedar&unscoped_q=cedar

PR #833, opened 2019-08-24
Upgrade to pip 19

cjolowicz:
This PR upgrades [pip](https://pypi.org/project/pip/) to version 19.x. This upgrade is the first step to provide support for `pyproject.toml`-based projects ([PEP 517](https://www.python.org/dev/peps/pep-0517/)), such as those managed by [poetry](https://poetry.eustace.io) and [flit](https://github.com/takluyver/flit).

- Upgrade to pip `19.2.3`, for projects using Python 2.7 and >= 3.5.
- Upgrade to pip `19.1.1`, for Python 3.4 projects. 

Note that pip 19.2 dropped Python 3.4 support, hence the version difference.

Patch the vendorized `pip-pop` package, which is used by the `pip-uninstall`, `gdal`, and `pylibmc` steps. It relies on internal pip modules changed by the update.

Closes #740
See #804

Follow-ups:

- #834 (Support pyproject.toml)
- #835 (Support poetry.lock)

---

CI is broken for PRs from another repository, see [this comment](https://github.com/heroku/heroku-buildpack-python/pull/832#issuecomment-524441204). Please check the corresponding PR in my fork to verify CI:

- cjolowicz/heroku-buildpack-python#13

PR #832, opened 2019-08-23
Fix typo in command to log pip-diff failure

cjolowicz:
Due to a typo, `mount(8)` is invoked instead of `mcount` from heroku/buildpack-stdlib.

The `pip-diff` tool from `vendor/pip-pop` is used to determine stale requirements. When `pip-diff` encounters an unexpected failure, a count should be logged using `mcount`.

PR #834, opened 2019-08-24
Support pyproject.toml

### LINKED ISSUE #796: Support for Poetry

cjolowicz:
Support [PEP 517](https://www.python.org/dev/peps/pep-0517/) package builds using the `pyproject.toml` file. This adds support for build backends other than setuptools, such as [poetry](https://poetry.eustace.io/) and [flit](https://github.com/takluyver/flit). This PR is a follow-up to #833.

There is not much to do here, because [pip](https://pypi.org/project/pip/) 19 already implements PEP 517-based builds. 

- Detect Python projects having only pyproject.toml and no requirements.txt or Pipfile.
- Avoid editable installs, which are not compatible with pyproject.toml-style projects.
- Add test cases for pyproject.toml-style projects managed by poetry and flit.

Note that while this PR does allow deploying poetry-based projects to Heroku, it does not use the pinned requirements from the `poetry.lock` file. Implementing support for poetry.lock is build backend-specific, and therefore left to be addressed by a follow-up PR.

Closes #796 

---

CI is broken for PRs from another repository, see [this comment](https://github.com/heroku/heroku-buildpack-python/pull/832#issuecomment-524441204). Please check the corresponding PR in my fork to verify CI:

- https://github.com/cjolowicz/heroku-buildpack-python/pull/6


cjolowicz:
It would be good to add a test case for a pyproject.toml using setuptools as the build backend end. There could be a problem here because the buildpack uses a vendorized setuptools pinned at 39.0.1.

cjolowicz:
Rebased onto master ([`v157`](https://github.com/heroku/heroku-buildpack-python/releases/tag/v157)).

cjolowicz:
Rebased onto master (af860be62afc610aa24e6a1556ab21279d3f5d0b)

cjolowicz:
Rebased onto master (93a620e6edacde750d32af6e3c8c3ad5e6800a40)

licht1stein:
Hi guys! Could you please add a comment on status of this and #796 Support for Poetry?

bfontaine:
Has there been any progress on this issue? Right now the build aborts if one has neither `requirements.txt`, `setup.py`, nor `Pipfile`. I have none of these, but I have a `pyproject.toml`, and I use [`python-poetry-buildpack`](https://elements.heroku.com/buildpacks/moneymeets/python-poetry-buildpack) alongside this buildpack.

edmorley (commented):
Hi! Sorry for the delay in replying here. We're in the process of moving to Cloud Native Buildpacks, so all development effort is focused there at the moment, and eventually they will replace the current buildpacks. As such, we're only making minimal changes to the current buildpacks for now. 

If you'd like to follow along with the development, our new CNB framework is here (which is now mostly complete):
https://github.com/heroku/libcnb.rs

And the Python CNB will (once it's beyond a local prototype), live here:
https://github.com/heroku/buildpacks-python
(My work in progress CNB has handling for `pyproject.toml` if it's any consolation!)

For more about CNBs, see:
https://buildpacks.io/

bin/compile
```diff
@@ -269,8 +269,13 @@ mtime "pip.uninstall.time" "${start}"
...
 
-if [ ! -f requirements.txt ] && [ ! -f Pipfile ]; then
-  echo "-e ." > requirements.txt
+if [ ! -f requirements.txt ] && [ ! -f Pipfile ] ; then
+    if [ -f pyproject.toml ] ; then
+        # Editable installs are not supported for pyproject.toml-style projects.
+        echo "." > requirements.txt
+    else
+        echo "-e ." > requirements.txt
+    fi
```
codingjoe, line 278: I don't think this change is needed. Editable installations are supported.

codingjoe (commented):
Hi @cjolowicz, thanks for the contribution. Let's try to push this over the finish line. I'd really love to use the modern project setup on Heroku.

The only thing needed to make this work, is the change in `bin/detect`.
Took make things easier for the maintainers, I'd suggest dropping all other changes.

I'd also recommend updating the README file. It should probably mention `pyproject.toml` support.

codingjoe:
# For everyone who needs a quick fix: Adding an empty `setup.py` file, is a good workaround.

dhdaines:
It would be nice if whatever eventually supports `pyproject.toml` could also support VCS-based versioning as mentioned in #1563

PR #835, opened 2019-08-28
Support poetry.lock

### LINKED ISSUE #796: Support for Poetry

cjolowicz:
Export pinned requirements from `poetry.lock` to `requirements.txt`. This PR is a follow-up to #833 and #834.

If a `poetry.lock` file is present, use the Poetry CLI to export the pinned requirements to a `requirements.txt` file. If the project already contains a `requirements.txt`, use that and ignore `poetry.lock`.

Install `pyproject.toml`-based projects using `pip install .`, instead of creating a `requirements.txt` containing `.`. The `requirements.txt` file is needed to handle pinned versions extracted from the `poetry.lock` file.

Closes #796 

---

CI is broken for PRs from another repository, see [this comment](https://github.com/heroku/heroku-buildpack-python/pull/832#issuecomment-524441204). Please check the corresponding PR in my fork to verify CI:

- cjolowicz/heroku-buildpack-python#7

---

**Why is this necessary at all, after supporting PEP 517 installations?**

Note that supporting `pyproject.toml`-based installations is not enough to handle pinned requirements: When pip installs a `pyproject.toml`-style project using the process described in [PEP 517](https://www.python.org/dev/peps/pep-0517/), it only uses Poetry to build a wheel. The wheel contains the version constraints from `pyproject.toml`, not the pinned versions from `poetry.lock`. See sdispater/poetry#537

**Why not use Poetry to install?**

Poetry is not used to install the project because it does not clean up stale requirements. This means that requirements need to be exported anyway, for the `pip-uninstall` step. See sdispater/poetry#648

**Why use Poetry 1.0.0b1 instead of the version specified in pyproject.toml?**

Since we only use Poetry to export a `requirements.txt` file, ignore the Poetry version specified in `pyproject.toml`. Install a pre-release of `1.0.0` because the export command is not available before `1.0.0a0`.

*Update*: Poetry 1.0.0 has been released. Also, the version specified in `pyproject.toml` is a minimum version, so it would not make sense to install it anyway.

**Why not append `.` to the exported requirements file?**

Note that we cannot simply append `.` to the exported `requirements.txt` file. Currently pip requires that either all requirements have a hash or none. Including `.` would thus force us to omit hashes for all requirements. See pypa/pip#4995

**Test case**

Add a test to check that pinned requirements are honored when `poetry.lock` is present. Request an outdated version of `marshmallow` in `poetry.lock` (`3.0.0`). Check that the outdated version is installed, rather than a newer version also satisfying the version constraint in `pyproject.toml` (`^3.0.0`).

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -269,8 +269,13 @@ mtime "pip.uninstall.time" "${start}"
 # This allows for people to ship a setup.py application to Heroku
 # (which is rare, but I vouch that it should work!)
 
-if [ ! -f requirements.txt ] && [ ! -f Pipfile ]; then
-  echo "-e ." > requirements.txt
+if [ ! -f requirements.txt ] && [ ! -f Pipfile ] ; then
+    if [ -f pyproject.toml ] ; then
+        # Editable installs are not supported for pyproject.toml-style projects.
+        echo "." > requirements.txt
+    else
+        echo "-e ." > requirements.txt
+    fi
 fi
 
 # Fix egg-links.
--- a/bin/compile
+++ b/bin/compile
@@ -268,14 +268,8 @@ mtime "pip.uninstall.time" "${start}"
 # If no requirements.txt file given, assume `setup.py develop` is intended.
 # This allows for people to ship a setup.py application to Heroku
 # (which is rare, but I vouch that it should work!)
-
-if [ ! -f requirements.txt ] && [ ! -f Pipfile ] ; then
-    if [ -f pyproject.toml ] ; then
-        # Editable installs are not supported for pyproject.toml-style projects.
-        echo "." > requirements.txt
-    else
-        echo "-e ." > requirements.txt
-    fi
+if [ ! -f requirements.txt ] && [ ! -f Pipfile ] && [ ! -f pyproject.toml ]; then
+    echo "-e ." > requirements.txt
 fi
 
 # Fix egg-links.
--- a/bin/compile
+++ b/bin/compile
@@ -257,6 +257,10 @@ mtime "python.install.time" "${start}"
 # shellcheck source=bin/steps/pipenv
 source "$BIN_DIR/steps/pipenv"
 
+# Export requirements.txt from poetry.lock, if present.
+# shellcheck source=bin/steps/poetry
+source "$BIN_DIR/steps/poetry"
+
 # Uninstall removed dependencies with Pip.
 # The buildpack will automatically remove any declared dependencies (in requirements.txt)
 # that were explicitly removed. This machinery is a bit complex, but it is not complicated.
--- a/bin/detect
+++ b/bin/detect
@@ -15,7 +15,7 @@
 BUILD_DIR=$1
 
 # Exit early if app is clearly not Python.
-if [ ! -f "$BUILD_DIR/requirements.txt" ] && [ ! -f "$BUILD_DIR/setup.py" ] && [ ! -f "$BUILD_DIR/Pipfile" ]; then
+if [ ! -f "$BUILD_DIR/requirements.txt" ] && [ ! -f "$BUILD_DIR/setup.py" ] && [ ! -f "$BUILD_DIR/Pipfile" ] && [ ! -f "$BUILD_DIR/pyproject.toml" ]; then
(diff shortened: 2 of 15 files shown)


cjolowicz:
Rebased onto master ([`v157`](https://github.com/heroku/heroku-buildpack-python/releases/tag/v157)).

- Tests were moved from `test/run` to `test/run-features`

cjolowicz:
Rebased onto master (af860be62afc610aa24e6a1556ab21279d3f5d0b)

cjolowicz:
Poetry 1.0.0b2 was released on 2019-10-11:

https://github.com/sdispater/poetry/releases/tag/1.0.0b2

cjolowicz:
Poetry 1.0.0b3 was released on 2019-10-25:

https://github.com/sdispater/poetry/releases/tag/1.0.0b3

cjolowicz:
Poetry 1.0.0b4 was released on 2019-11-08:

https://github.com/sdispater/poetry/releases/tag/1.0.0b4

cjolowicz:
- Rebased onto master (93a620e6edacde750d32af6e3c8c3ad5e6800a40)
- Upgrade to Poetry 1.0.0b7 (03e27237193405f4ce017fe107b7ea7154aa5883)
- Minor cleanup (3998f12508bfe2245da76e3b0b10b8b0421180b1)

cjolowicz:
@CaseyFaist : Any chance of getting this series of PRs merged by the end of 2019?

orokusaki:
Poetry 1.0.0b8 was released a few days back (fixes at least one bug)

adamkgoldfarb:
Poetry 1.0 was officially released today, formalizing the export command.

cedricbonhomme:
Could someone give us an idea of when this PR will be merged ?
Thank you.

zyv:
A stop-gap solution: https://elements.heroku.com/buildpacks/moneymeets/python-poetry-buildpack

yunti:
What does heroku currently default to (prior to this PR being merged) if their is a `pyproject.toml` file and a `requirement.txt` file (that has previously been generated with `poetry export ..`)? Does it currently pick up and use that hashed and fully resolved `requirement.txt`?

zyv:
@yunti this PR is nowhere near to be merged, so if both files are present currently Heroku defaults to using `requirements.txt` - that’s what we did before writing our own overlay buildpack.

edmorley (commented):
Hi! Thank you for the PR. We're in the process of moving to Cloud Native Buildpacks, so all development effort is focused there at the moment, and eventually they will replace the current buildpacks. As such, we're only making minimal changes to the current buildpacks for now. 

For more information, see:
https://github.com/heroku/heroku-buildpack-python/pull/834#pullrequestreview-952972169

Strokes-LastNight-Logged-Keystrokes:
If you're really keen to use tarball.gz for encryption, first thing you need to do to support toml.pythonproject compliance is to geolocate hidden controllers in the index using C## pointers. So those with UNIX distro can make the following initial assumptions:

14_anchor_size.c
c utf-8 [unix] 86% In: 19%: 4? E

#include <stdio.h>
#include <stdlib.h>

#define SIZE 100

//size necessary
void print_elements (int *v)
for (int i=0; i< ???; ++i)
      print f("%d/n", v[i]);

it l{fncn%s}/n=4:
[length, percentage], assert global var "variance" q> l{fncn%s}/n;
catch funcn?
map fncn, write to self where {P, B}F ;

Define "float" pointer
# include <stdio.h>

int main( )
{ [int props, ext props] int nb;
     int margin * q(ptr) = {"float"ptr}.it.count*2.
      while_instance{"float"ptr} [int nb; int hexdec]>=2, assert state is 0.6325/0+ 0.7745/1> printf("%d, %"float" pointer); useState.asyn.io

edmorley:
@cjolowicz Thank you for the PR! A lot has changed since it was opened - amongst other things the Poetry bug mentioned in the OP was fixed some time ago, so it's now viable to use Poetry to install the dependencies rather than exporting to a requirements file and then using pip to install that.

I've just added Poetry support in #1682 :-)

PR #840, opened 2019-09-11
Move get-pip utility to s3

CaseyFaist:
The Python Buildpack uses the community-supported utility https://github.com/pypa/get-pip to install pip during builds.

This is a Python script that contains a large binary download of pip (previously 9.0.2) to install the pip version requested by the `PIP_UPDATE` variable here: https://github.com/heroku/heroku-buildpack-python/blob/master/bin/steps/python#L110

This change is to remove it from the buildpack and instead place it on S3. The reasoning behind this is that binaries belong on S3 versus in the buildpack, and updates to this file can be rolled out quickly and easily. 

The buildpack now points to an updated get-pip script with this change.

This will resolve issues building Python 3.7.4 and 3.6.9 on all stacks. 

Also: Makefile updates to correctly point to the heroku/cedar docker hub images and used for local testing.


bin/steps/python
```diff
@@ -97,6 +97,15 @@ if [ ! "$SKIP_INSTALL" ]; then
   hash -r
 fi
 
+# Heroku uses the get-pip utility maintained by the Python community to vendor Pip.
+# https://github.com/pypa/get-pip
+GETPIP="https://lang-python.s3.amazonaws.com/etc/get-pip.py"
```
jmorrell: Could you add the SHA to the filename here? and check the SHA after downloading?

```
GETPIP="https://lang-python.s3.amazonaws.com/etc/get-pip-${SHA}.py"
```

This ensures that the bytes you download are the same ones you are expecting, and that when you need to update it, that is a distinct PR that goes through code review before getting released.

bin/steps/python
```diff
@@ -97,6 +97,22 @@ if [ ! "$SKIP_INSTALL" ]; then
...
+
+# Grab SHA from get-pip
+CHECK_SHA="$(curl -s "$GETPIP" | sha256sum | cut -d ' ' -f 1)"
```
jmorrell: There's no need to download this twice. You can check the SHA of the downloaded file.

bin/steps/python
```diff
@@ -97,6 +97,22 @@ if [ ! "$SKIP_INSTALL" ]; then
   hash -r
 fi
 
+# Heroku uses the get-pip utility maintained by the Python community to vendor Pip.
+# https://github.com/pypa/get-pip
+GETPIP="https://lang-python.s3.amazonaws.com/etc/get-pip.py"
```
jmorrell: The SHA should also go in the filename, or you may find there are race conditions when you try to update it.

Issue #824, opened 2019-06-07
Pipenv installs dependencies twice in test environment (redundant)


alexandre-paroissien:
Hi it seems that pipenv is running twice in test environment, once for dependencies (pipenv's "packages") and once for test dependencies (pipenv's "packages" + "dev-packages")

https://github.com/heroku/heroku-buildpack-python/blob/master/bin/steps/pipenv

```
-----> Postgresql done

-----> Python app detected

-----> Installing dependencies with Pipenv 2018.5.18…

   Installing dependencies from Pipfile.lock (7877d8)…

-----> Installing test dependencies…

   Installing dependencies from Pipfile.lock (7877d8)…

.-----> Running test-setup command python manage.py migrate --settings=config.heroku_ci_test_settings..
```

The issue happens when running the tests with Heroku CI, the test setup will run "pipenv install" first and then "pipenv install --dev" which is redundant (reinstalls the packages from the first command) and running these two commands is equivalent to running "pipenv install --dev" only

```
   Install all dependencies for a project (including dev):
   $ pipenv install --dev
```

https://github.com/pypa/pipenv

I was able to reproduce this issue with "heroku/python" buildpack, as well as "https://github.com/heroku/heroku-buildpack-python.git" buildpack

But the issue is not when the buildpack is used by the dyno (the dependencies are installed only once during the build which is correct)

How can I only install pipenv's "packages" in test environment?

A quick fix for my issue was to fork the buildpack, comment the unwanted dev-dependencies, and to use it as the test buildpack in app.json

https://github.com/jb-engine/heroku-buildpack-python/commit/22657fd2f9320d4aaf515d46a718df9e59f20d0a

So in my case I wanna run only "pipenv install" for the tests, I am thinking there should be an option to only install main dependencies for those who want it (but by default it should install dev dependencies for everyone else for compatibility with current behaviour, but it should not run both commands because it is redundant and can be quite slow)

alexandre-paroissien:
@CaseyFaist Hello there, I just tested again with "heroku/python" release as well as latest master branch release (ff42f31) 
Unfortunately I am still getting the dependencies installed twice in Heroku Tests as you can see here:

![image](https://user-images.githubusercontent.com/34052139/65035463-7346d880-d973-11e9-9026-0016806d6940.png)

`pipenv install --dev` would give exactly the same result as `pipenv install; pipenv install --dev` except it would save a lot of computing time and bandwidth on the Test Dynos

[...]

PR #851, opened 2019-09-24
Fix pipenv install twice in CI

alexandre-paroissien:
I noticed I was getting pipenv dependencies installed twice in Heroku CI tests ( #824 ) which is redundant and time consuming during the tests as you can see here:

image

As pipenv documentation shows it, installing dev dependencies includes all the main dependencies:

```
   Install all dependencies for a project (including dev):
   $ pipenv install --dev
```

https://github.com/pypa/pipenv

So made this PR and I tested this Heroku buildpack with Heroku CI test (it installs dev dependencies as currently) and Dyno build (it installs main dependencies as currently)

I moved the if-block responsible for test (main+dev) dependencies above the "if-block" for dyno build (main) dependencies which I changed into a "elif-block" to avoid the redundant install.

So it is the same behaviour as before except tests will be faster thanks to removed redundancy:

Tests in CI:
image

Dyno Build:
image

Let me know if I should change or improve anything

### DIFF
--- a/bin/steps/pipenv
+++ b/bin/steps/pipenv
@@ -60,10 +60,16 @@ if [ ! "$SKIP_PIPENV_INSTALL" ]; then
         # avoid this eager behavior.
         /app/.heroku/python/bin/pip install pipenv==$PIPENV_VERSION --upgrade --upgrade-strategy only-if-needed &> /dev/null
 
+        # Install the test dependencies, for CI.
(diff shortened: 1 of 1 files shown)


CaseyFaist:
Grabbed this and superceded with #861 to run tests, thank you @alexandre-paroissien 🎉

alexandre-paroissien:
You're welcome @CaseyFaist

PR #858, opened 2019-09-27
Fix sqlite

CaseyFaist:
Rebased from #812 for further work + upstreaming - see other pr for full context on approach


builds/cedar-14.Dockerfile
```diff
@@ -7,7 +7,7 @@ ENV WORKSPACE_DIR="/app/builds" \
     DEBIAN_FRONTEND=noninteractive \
     STACK="cedar-14"
 
-RUN apt-get update && apt-get install -y python-pip && rm -rf /var/lib/apt/lists/*
+RUN apt-get update && apt-get install -y python-pip libsqlite3-dev && rm -rf /var/lib/apt/lists/*
```
dzuelke: This should be done inside the python formula, not globally (`apt-get update && apt-get install -y libsqlite3-dev`). See https://github.com/heroku/heroku-buildpack-php/blob/343fa959d4c195beb7febe1996340624baf713b9/support/build/php#L59-L71 for how to do this (the stack specific additions there are of course not relevant here)

builds/heroku-16.Dockerfile
```diff
@@ -7,7 +7,7 @@ ENV WORKSPACE_DIR="/app/builds" \
     DEBIAN_FRONTEND=noninteractive \
     STACK="heroku-16"
 
-RUN apt-get update && apt-get install -y python-pip && rm -rf /var/lib/apt/lists/*
+RUN apt-get update && apt-get install -y python-pip libsqlite3-dev && rm -rf /var/lib/apt/lists/*
```
dzuelke: This should be done inside the python formula, not globally (`apt-get update && apt-get install -y libsqlite3-dev`). See https://github.com/heroku/heroku-buildpack-php/blob/343fa959d4c195beb7febe1996340624baf713b9/support/build/php#L59-L71 for how to do this (the stack specific additions there are of course not relevant here)

dzuelke: (also, this is missing from `heroku-18.Dockerfile` right now)

builds/cedar-14.Dockerfile
CaseyFaist: thanks for sharing an example on how - can you say more about why so I understand?

dzuelke:
So right now this won't allow `pip install`s of packages that need the headers.

To achieve that, we need to, roughly, do this:

1. `apt-get install libsqlite3-dev` in the Python formula (see https://github.com/heroku/heroku-buildpack-php/blob/343fa959d4c195beb7febe1996340624baf713b9/support/build/php#L59-L71 for a generic way of doing this)
1. ensure `realpath` is also `apt-get install`ed on cedar-14 in the Python formula (see previous point's link for the pattern)
1. `mkdir -p ${OUT_PREFIX}/include ${OUT_PREFIX}/lib/x86_64-linux-gnu`
1. `cp /usr/include/sqlite3*.h ${OUT_PREFIX}/include` 
1. `ln -fs $(realpath /usr/lib/x86_64-linux-gnu/libsqlite3.so) ${OUT_PREFIX}/lib/x86_64-linux-gnu/libsqlite3.so`

This will ensure that the libsqlite3 headers, as well as a symlink for the generic `.so` library (which gets resolved when compiled against) are included in the built Python tarball

Then, in `bin/compile`, just before `pip install`, we need to point the compiler infrastructure at those additional locations:

1. `export CFLAGS="-I/tmp/libsqlite3-dev/usr/include ${CFLAGS:-}"`
1. `export LDFLAGS="-L/tmp/libsqlite3-dev/usr/lib/x86_64-linux-gnu ${LDFLAGS:-}"`

This should then allow building e.g. a `pysqlite` dependency in `requirements.txt`.

builds/cedar-14.Dockerfile
CaseyFaist: Reason: apt-buildpack (suggested previous solution) won't load headers, and thus this would not allow users of pysqlite or other utilities to use sqlite on runtime.

CaseyFaist:
Capturing for future reference:

The reason for the shift in approach is that the [apt-buildpack](https://github.com/heroku/heroku-buildpack-apt) won't load headers properly, and thus would not allow users of pysqlite or other utilities to use sqlite on runtime. We're opting to preserve this functionality instead, with added complexity as the tradeoff 🎉

CaseyFaist:
Also important: the `c` and `ld` flags will **not** be updated in this pr, but a future one when this functionality is flipped on. That's a potentially breaking change, so slicing this up there unblocks the rest of this work without causing breakage.

PR #860, opened 2019-09-30
Test staged binaries

CaseyFaist:
Travis configuration changes

Matrix should now test both the production VENDOR_URL source of Python binaries and the staging python binaries

This PR requires some binary backfill to succeed, so first test run is to ensure the Travis matrix generates tests correctly


CaseyFaist:
These are false positives :( need to pass the vendor_url into the docker environments directly, not just the travis environment

CaseyFaist:
These are still false positives - no 3.4 binary exists in staging, for example, so TESTRUN=run-versions should be failing.

CaseyFaist:
[BOOYA TEST FAILS](https://travis-ci.org/heroku/heroku-buildpack-python/builds/595214305?utm_source=github_status&utm_medium=notification)

PR #861, opened 2019-09-30
Fix pipenv install twice with CI

CaseyFaist:
by @alexandre-paroissien 🌮 ❤️ 🌮 

Supercedes #851 to run secure tests


bin/steps/pipenv
```diff
@@ -60,10 +60,16 @@ if [ ! "$SKIP_PIPENV_INSTALL" ]; then
         # avoid this eager behavior.
         /app/.heroku/python/bin/pip install pipenv==$PIPENV_VERSION --upgrade --upgrade-strategy only-if-needed &> /dev/null
 
+        # Install the test dependencies, for CI.
+        if [ "$INSTALL_TEST" ]; then
```
jkutner: I wonder if `PIP_INSTALL_TEST` would be a better env var name?

jkutner: Oh nevermind, i see this was already used as an env var name

PR #865, opened 2019-10-02
Add env check to disable sqlite3 install

drgarcia1986:
Hi, guys, I don't know if it's the better way to do that but, I would like a way to disable sqlite3 installation. I just follow the rule to skip _django collect static_.

### DIFF
--- a/bin/steps/sqlite3
+++ b/bin/steps/sqlite3
@@ -70,24 +70,26 @@ sqlite3_install() {
 }
 
 buildpack_sqlite3_install() {
-  sqlite3_version
-  HEROKU_PYTHON_DIR="$BUILD_DIR/.heroku/python"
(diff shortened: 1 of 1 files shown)


CaseyFaist:
Hey - Thanks @drgarcia1986 for the PR, but unfortunately, more changes and engineering work would be required to make this work. You can check out #858 for context, but we not only vendor the headers in, but the Python runtime is built with sqlite - I'd need to build new python binaries as well for a feature like this to work. Not out of the question - but out of scope for now.

Do leave me an issue though with this feature request so I can come back to it though 😄

oTree-org:
Hi, I'd also be interested in a way to skip the SQLite installation step. It seems to be the most time-consuming step of our builds and our projects don't use SQLite. Faster builds would really help as we have many apps and do a lot of rapid iteration.

PR #867, opened 2019-10-07
Presence of pyproject.toml (PEP 518) triggers python detection.

DanLipsitt:
Projects using the `pyproject.toml` file as specified in [PEP 518](https://www.python.org/dev/peps/pep-0518/) do not require a `setup.py` or `requirements.txt` file. Pipenv doesn't use PEP 518, so a `Pipfile` won't be present either. Therefore these will not be detected as Python projects. This patch implements detection using `pyproject.toml`. This includes (and is motivated by) projects using the [Poetry](https://poetry.eustace.io) packaging tool.

There are other open tickets regarding Poetry, such as #801. This patch doesn't implement full Poetry support, just Python detection.


CaseyFaist:
Hi there @DanLipsitt - thanks for the pull request! ❤️ 🌮 💯 This change has some prereqs (which you've seen), so while I appreciate the contribution, I'd like to punt on this until those changes are closer to release. I'm going to take the opportunity to revisit how detection works and which files we check for.

Issue #852, opened 2019-09-26
Support for vendoring our own versions of GEO libraries


thclark:
At present, the buildpack, when installing GEO libraries, uses the ``${VENDOR_URL}`` environment variable to determine which libraries to install.

Those of us who need to build our own geo libraries can't do so in the build phase, because gdal alone takes c.1hr to build, and so we hit Heroku's timeout. We have to vendor our own libraries.

This would be extremely simple if the geo libraries were installed from ``${GEO_VENDOR_URL}``, an envvar we could set (whose default value is otherwise the same as ``${VENDOR_URL}``). At present however, because all the vendored libraries are using the same origin, it's a much more complicated job involving either an additional/alternative buildpack or at least a postbuild hook.

Here's all the different uses of ``VENDOR_URL`` in the buildpack:
![Screenshot 2019-09-26 at 14 36 23](https://user-images.githubusercontent.com/7223170/65693647-8d1ba600-e06c-11e9-8b4e-b39de85a0bc3.png)

I believe the modification would be made [here](https://github.com/heroku/heroku-buildpack-python/blob/ff42f318a06b80873331bdd62a7e335573a35534/bin/compile#L42) and is as simple as:
```
VENDOR_URL="https://lang-python.s3.amazonaws.com/$STACK"
if [[ -n ${BUILDPACK_VENDOR_URL:-} ]]; then
    VENDOR_URL="$BUILDPACK_VENDOR_URL"
fi

GEO_VENDOR_URL="$VENDOR_URL"
if [[ -n ${GEO_VENDOR_URL:-} ]]; then
    VENDOR_URL="$GEO_VENDOR_URL"
fi

export GEO_VENDOR_URL
export VENDOR_URL
```

(ofc the references to it in the build steps would also need updating).

Use cases:
- you need plugins against which default vendored versions aren't linked, like Google's libkml
- you need a different version of GDAL to the default vendored version (at the time of writing, latest is 2.4.1, vendored is 2.2.1)

**Would you accept a pull request for this, which updated the compile script, the build steps, and the docs accordingly?**

CaseyFaist:
Hi @thclark!

We will be deprecating support for in-buildpack gdal/geos installation, so while I appreciate the issue documentation and the offer to PR  ❤️ I won't be able to accept a pull request on this issue.

[...]

PR #869, opened 2019-10-08
pyproject.toml revisited

DanLipsitt:
Apologies for being stubborn, but I'm unable to comment on my previous closed pull request and I think @CaseyFaist may have misunderstood the intent of #867. This is a stand-alone patch with no prerequisites, and I think it has value independent of what appears to be in other tickets. It does not attempt to integrate with projects that use Poetry, it just allows projects to be recognized as using Python in a way that should not interfere with any other functionality. I'm glad that other detection changes may be coming down the pike, but this change would be beneficial until those changes take place or in case they do not get made. In particular, python projects that use pyproject.toml on [GitLab's AutoDevOps system](https://docs.gitlab.com/ee/topics/autodevops/#auto-build-using-heroku-buildpacks) fail immediately because they are not recognized as Python.

Thank you for your time. I understand if this patch doesn't fit into your plans. I just want to make sure you understand what I'm trying to do.


CaseyFaist:
Hi again @DanLipsitt - no worries, thanks for coming back with more context ❤️ 🌮 In the future, if it looks like a maintainer has misunderstood the intent or the problem you're trying to address, could you open an issue to capture that context? You can still describe your recommended fix, and ask if the maintainer can accept a pr to address - [this issue](https://github.com/heroku/heroku-buildpack-python/issues/852) is an awesome example of that.

So, my confusion here is that the version of Pip we're currently on does not support pyproject.toml files (pip==9.0.3). For this buildpack to successfully build, an app needs a requirements.txt file, a pipfile, or a setup.py file. It would not be a good experience to recognize applications with none of those files as Python, as their dependencies won't be installed (although they'll be assigned the default Python 3.6 runtime).

The goal of this buildpack is not to provide an empty python shell, but to provide a python environment to specifications - but my concern is this pr would cause the former behavior.

Pyproject.toml support is planned, but I'm not sure how this change can stand independently of the other work involved. Until the buildpack takes the proper action when presented with a pyproject.toml, the absence of the other files *should* cause a failure.

CaseyFaist:
Closing this due to no further discussion, but please do open an issue to discuss further if I've still misunderstood the goal and impacts of this work 👍

PR #870, opened 2019-10-08
SQLITE: Py3 formula update

CaseyFaist:
This is the first of several PRs updating the SQLITE3 dynamic linking during Python binary creation. Broken out from #858 for better release flow.


danielleadams:
@CaseyFaist Looks like I can approve when it is still a draft! Not sure what will happen with Compliance when it becomes a PR... may have to go through it again

PR #868, opened 2019-10-07
Refactor: $BUILD_DIR

CaseyFaist:
In most places (aside from possibly simlinks) /app/ is the same as the $BUILD_DIR var used throughout the buildpack. 

@duanehutchins refactored to reflect this, making the buildpack cleaner and more consistent  🎉

This pr supercedes #761 so I can run tests with creds

Issue #873, opened 2019-10-09
ln: failed to create symbolic link '/tmp/build/.heroku/python': No such file or directory


wckao:
https://github.com/heroku/heroku-buildpack-python/commit/05e29c74bcc4b73a0ab2690f767a53bfdbd5de13#r35419023

wckao:
[deploy_log.txt](https://github.com/heroku/heroku-buildpack-python/files/3705539/deploy_log.txt)

Here is the complete log

jacobwegner:
I think this recent change is causing the failure:

https://github.com/heroku/heroku-buildpack-python/blob/1721fa3703ca85afecb09656299866c036461a66/bin/compile#L108

```patch
-mkdir -p "BUILD_DIR/.heroku"
+mkdir -p "$BUILD_DIR/.heroku"
```

cw-intellineers:
I have the same issue. Will there be a hotfix in the near future? We are unable to deploy our changes otherwise.

vinicyusmacedo:
Hey, I would like to grab this issue.

PR #874, opened 2019-10-09
Revert "Refactor: $BUILD_DIR"

dzuelke:
Reverts heroku/heroku-buildpack-python#868

The reason this breaks is that `$BUILD_DIR` is `/tmp/…`, so all the library includes etc from builds now point to that.

Python is not relocatable, so this essentially gets "hardcoded" into the built slug.

That's why a symlink is set up from `/app/.heroku/python` to `$BUILD_DIR/.heroku/python`: so the hard-coding of `/app/.heroku/python` works at deploy time, and at runtime.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 # Master
 
-- Refactor: use variable rather than hardcoded /app
 - Bug fix: pipenv no longer installs twice on CI
 
 --------------------------------------------------------------------------------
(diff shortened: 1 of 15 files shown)


vinicyusmacedo:
hey @dzuelke, according to https://github.com/heroku/heroku-buildpack-python/issues/873#issuecomment-539835822, there's a dollar sign missing which might be breaking the mkdir command. Maybe it would be better to put it there instead of removing it at all?

dzuelke:
Revert first, fix investigation later.

As far as I can tell, that entire previous PR is not suitable for merging, since it will likely break certain builds (the `/app` hardcoding exists for a reason, since Python is not relocatable).

PR #875, opened 2019-10-09
Fixed BUILD_DIR bug introduced in #868

duanehutchins:
Fixes bug introduced in #868 by changing BUILD_DIR INTO $BUILD_DIR


dzuelke:
Revert was done in #874, and the whole change needs more evaluating anyway.

PR #882, opened 2019-10-14
Staging binary controller env var

CaseyFaist:
Initially I tried to use the default $BUILDPACK_VENDOR_URL provided, but it wasn't handling the $STACK var appropriately - this is a quick fix to concatenate the supplied staging binary with the current STACK provided to the build environment.

Related: #860

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -42,6 +42,8 @@ export BUILD_DIR CACHE_DIR ENV_DIR
 VENDOR_URL="https://lang-python.s3.amazonaws.com/$STACK"
 if [[ -n ${BUILDPACK_VENDOR_URL:-} ]]; then
     VENDOR_URL="$BUILDPACK_VENDOR_URL"
+elif [[ -n ${USE_STAGING_BINARIES} ]]; then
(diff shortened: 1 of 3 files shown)


test/run-versions
```diff
@@ -4,8 +4,6 @@
 # shellcheck source=bin/default_pythons
 source "bin/default_pythons"
 
-assertCaptured "Installing SQLite3"
```
CaseyFaist: This is cleanup - this line should not exist outside of a test

bin/compile
```diff
@@ -42,6 +42,8 @@ export BUILD_DIR CACHE_DIR ENV_DIR
 VENDOR_URL="https://lang-python.s3.amazonaws.com/$STACK"
 if [[ -n ${BUILDPACK_VENDOR_URL:-} ]]; then
     VENDOR_URL="$BUILDPACK_VENDOR_URL"
+elif [[ -n ${USE_STAGING_BINARIES} ]]; then
```
CaseyFaist: `:-` is defining a blank default value, I think. If so, it's not necessary

PR #889, opened 2019-10-21
Added a deprecation warning when using BUILD_WITH_GEO_LIBRARIES

KevinBrolly:
Added a deprecation warning informing users that the `BUILD_WITH_GEO_LIBRARIES` feature and associated Geo libraries are being deprecated and will be removed from the buildpack in the future.


CaseyFaist:
Rebased and superceded by #895 🎉

PR #891, opened 2019-10-22
Python 2.7.17

CaseyFaist:
This adds Python 2.7.17 support and sets it as the default Python 2.7 version.

- default-pythons updated
- appropriate test version bumped
- binaries generated


edmorley:
@CaseyFaist Hi! I spotted the build script added here doesn't make use of the base Python 2 script (`builds/runtimes/python2`) like the other scripts. Was this intentional, or because this PR raced with the landing of #881?

It's just that the script here (`builds/runtimes/python-2.7.17`) doesn't include the latest sqlite3 symlink parts from `python2` and instead uses `sqlite3_install` which invokes APT?

PR #897, opened 2019-10-25
Add support for Python 3.8 latest version

codingjoe:
If the pip lock file only specifies `3.8` and no bug fix version, it should use Python LATEST_38.

@CaseyFaist I found the bug, this should do the trick.


codingjoe:
@CaseyFaist is there any chance this bugfix will make it into master soonish?

CaseyFaist:
To be clear, we are not accepting pipenv feature requests at this time - however, this appears to be purely a bug fix, so looks good 👍 will test from privileged account next week.

codingjoe:
Yes, this is working for all version, but 3.8, so we can sneak this in 😉
Oh BTW, are there any plans to support poetry in the future?

jkutner:
@codingjoe thanks for the fix. I'm closing this, and merging it into #955 so that CI will pass (it requires a token so it won't pass on forked PRs)

codingjoe:
Thanks @jkutner !

PR #899, opened 2019-10-29
Use latest patch version

CaseyFaist:
This will allow users to specify a python major and minor branch to always use the latest patch available.

`python-X.X.*` is the acceptable format, where X.X are the major and minor versions.

Available versions for this tooling:
```
python-3.8.*
python-3.7.*
python-3.6.*
python-2.7.*
```


CaseyFaist:
Hopefully will get to come back around to this next week, I'm excited about this feature 🎉

bin/steps/python
```diff
@@ -4,6 +4,21 @@ set +e
 runtime-fixer runtime.txt
 PYTHON_VERSION=$(cat runtime.txt)
 
+if [[ $PYTHON_VERSION == *"*"* ]]; then
+  if [[ $PYTHON_VERSION == *"3.8"* ]]; then
```
edmorley, line 8: I think this would cause false positives - `$PYTHON_VERSION == *"3.8"*` will match against versions like `*.3.8` or `2.3.8*`

edmorley:
This is definitely something I'd like the buildpack to support (and as soon as possible!).

I've added some thoughts as to implementation approaches to #913 - closing this PR out since I'll open a new one once I start working on this :-)

PR #900, opened 2019-10-29
Resolve #819 -- Move shellcheck to GitHub action in favor of Travis-CI

### LINKED ISSUE #819: check out shellcheck action
https://github.com/marketplace/actions/shellcheck

codingjoe:
I saw that you are already using GitHub actions, so I figured why not start to move some other bits like shellcheck. This is particularly important since travis currently does not seem to work. Which means that shellcheck isn't even executed. Having it as a separate action will allow it to be executed in parallel and give feedback on GitHub for contributors.


CaseyFaist:
For some reason this is not triggering the test suite, which is required for code security - Pulling down locally to see if it's a user permission thing. If that's all it is, I'll schedule this for Monday release 👍

CaseyFaist:
Closing with resolved #929, scheduled for release Monday 🌮

PR #903, opened 2019-12-02
Update README python versions to match docs

barbeque:
The README and [the official docs page](https://devcenter.heroku.com/articles/python-support) disagree on which versions are supported. I assumed that the official docs were correct, so here is a quick PR to make these consistent. Thanks!


CaseyFaist:
Hi there! Thanks for the contribution ❤️ 

Could you:

- [ ] Add a Changelog update under the Master heading describing the change
- [ ] Update to match current docs

If you can take care of those action items, this can be scheduled for Monday or Thursday next week release 🎉

barbeque:
All set, thank you!

CaseyFaist:
Booya!! I'll take it from here 🎉 I'll need to run the PR from a privileged account, so you'll see this PR close, but your commits are the ones I'll be pushing up (with your name attached). Thanks for your contribution! ❤️

PR #904, opened 2019-12-02
Sqlite quickfix

CaseyFaist:
Quick version bump to address bug causing failed builds.

If this is affecting your build, see here for a temporary workaround:

https://status.heroku.com/incidents/1802


Resinderate:
Thanks for the quick fix 👍

nwjlyons:
Thanks you for the fix 🙏 

Just an FYI. In the linked issue https://status.heroku.com/incidents/1802 it says it only happened on the `heroku-16` stack, but I had the issue on the `heroku-18` stack while creating review apps.

PR #906, opened 2019-12-04
Allow other pipenv versions than 2018.5.18

Thomas-Mollard:
Following the issue https://github.com/heroku/heroku-buildpack-python/issues/786 and the PR https://github.com/heroku/heroku-buildpack-python/pull/820 that has not been merged (and is now out of date with the master branch), I am changing a bit the code of https://github.com/heroku/heroku-buildpack-python/pull/820 to allow other pipenv versions to be installed while keeping the default at `2018.5.18`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 # Master
 
+- Allow other Pipenv versions than 2018.5.18
 - Update requirements.txt builds to use Pip 20.0.2
 - Download get-pip.py to tmpdir instead of root dir
 
--- a/bin/steps/pipenv
+++ b/bin/steps/pipenv
@@ -52,7 +52,7 @@ if [ ! "$SKIP_PIPENV_INSTALL" ]; then
             mcount "buildvar.SLUGIFY_USES_TEXT_UNIDECODE"
         fi
 
-        export PIPENV_VERSION="2018.5.18"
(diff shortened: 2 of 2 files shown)


paulmelnikow:
@CaseyFaist Can this be merged? I'm getting stuck on #704 and this would provide a workaround.

CaseyFaist:
Hi there! Thank you for your contribution ❤️ however since it's currently unclear how supported the pipenv project is, no pipenv feature requests are being prioritized at this time. 

I am sorry about this frustration, however our version (2018.5.18) is only one version behind the current version (2018.11.26), and since pipenv vendors large chunks of pip into its source code, it's unclear which versions of pip the new version would be compatible with.

Also of note, the current pinned version of Pipenv would require users to uninstall and reinstall their local pipenv environments before being able to deploy again successfully to generate a valid pipfile. This is due to incompatibility with the lock files that is not autocorrected by pipenv's update function, and would lead to a lot of confusion for users attempting to update and expecting the updated pipfile to deploy successfully with the new version. 

Since this would require coordination between the release and all pipenv users, or would result in a temporarily opt-in upgrade (which, if anyone missed the communications around it, could lead to quite a lot of frustration), we're not looking at this update any time soon.

Mx-Glitter:
@CaseyFaist Can we revisit this when [Pipenv gets its new release](https://github.com/pypa/pipenv/issues/3369)?

PR #910, opened 2019-12-20
Fix runtime.txt generation when using pipenv

Thomas-Mollard:
The sentence in the README explaining how runtime.txt is automatically generated when using pipenv is not correct as it is referring to `python-version` which is not very clear (where is this field?) and not correct (it should be either `python_version` or `python_full_version`)


jkutner:
@Thomas-Mollard thanks for fixing this

PR #919, opened 2020-02-19
Upgrade to pip 20

CaseyFaist:
Cleanup, bumping forward to version 20, and running tests from my account (with credentials) - otherwise all work by @cjolowicz ❤️ 🎉 ❤️


Hybridx24:
Awesome! But is there a way to make this work with `pipenv`? When I push a package to Heroku that uses pipenv, it shows that I am still running pip v9

Hybridx24:
@CaseyFaist Awesome! But is there a way to make this work with pipenv? When I push a package to Heroku that uses pipenv, it shows that I am still running pip v9 (and packages like pymupdf fail to install)

Issue #909, opened 2019-12-11
Cache unnecessarily being cleared due to lack of cached SQLite3 version file


luyizhao:
Post #907, we are no longer caching a `.heroku/python-sqlite3-version` version file and instead always running `sqlite3_install "$BUILD_DIR/.heroku/python"`. The missing version file causes builds now to unnecessarily clear their caches.

Because we no longer cache a `.heroku/python-sqlite3-version` version file, we are clearing all `.heroku` caches in https://github.com/heroku/heroku-buildpack-python/blob/v162/bin/steps/python#L72 because 
```bash
if [ -d .heroku/python ] && [ ! -f .heroku/python-sqlite3-version ] && python_sqlite3_check "$PYTHON_VERSION"; 
```
in https://github.com/heroku/heroku-buildpack-python/blob/v162/bin/steps/python#L70 will now always evaluate to true (for builds with caches and for python versions requiring sqlite3, i.e. where `python_sqlite3_check` is true).

I am actually not sure if the cache clearing logic https://github.com/heroku/heroku-buildpack-python/blob/v162/bin/steps/python#L67-L73 (originally introduced in #713) is still needed. If it isn't, a straightforward fix would be to remove those lines of cache-clearing logic. (Otherwise, we could still maintain a cached `.heroku/python-sqlite3-version` version file as we did before prior to #907)

Relatedly, the following related few lines from https://github.com/heroku/heroku-buildpack-python/pull/713 now appear to be outdated and can be removed now as well:
- https://github.com/heroku/heroku-buildpack-python/blob/v162/bin/compile#L169-L170
- https://github.com/heroku/heroku-buildpack-python/blob/v162/bin/steps/sqlite3#L62-L65

sidmitra:
I'm wondering if this is the reason why builds were taking 25 mins on heroku CI. I had lxml, pandas, numpy and pip installing these was what was taking the bulk of the time. Our actual test run is only 3-4 mins whereas the test setup runs for ~25 mins.

```
...
       Running setup.py install for numpy: started
.           Running setup.py install for numpy: still running...
.           Running setup.py install for numpy: still running...
.           Running setup.py install for numpy: still running...
.           Running setup.py install for numpy: still running...
.           Running setup.py install for numpy: still running...
           Running setup.py install for numpy: finished with status 'done'
         Running setup.py install for pandas: started
.           Running setup.py install for pandas: still running...
.           Running setup.py install for pandas: still running...
[...]

PR #925, opened 2020-03-12
Fix cache clear on redeploy (no pip changes)

CaseyFaist:
While investigating #909, discovered that upgrading pip to v20 broke a legacy internal tool [pip-diff](https://github.com/heroku/heroku-buildpack-python/blob/master/vendor/pip-pop/pip-diff).

This PR updates the buildpack so now we:

1. Cache `requirements.txt` (heroku does not preserve git history after compiling the slug)
2. Check for cached `requirements.txt` 
3. Crudely check for any file change using `diff` if file is found
4. Clear the cache if requirements file has been updated (to be improved after bug is resolved)
5. OR clear if no cached requirements is found (for the first time applications deploy after this change)
6. Reinstall environment after changes
7. No longer use broken script to check for stale dependencies
8. Drop the offending clear-cache line from the SQLITE check - we now install latest sqlite from resource::list without clearing cache

Alternatives considered:
- Use Pip Hashing to hash the dependencies and cache/check for hash changes
Could not find a way that did not require users to hash on their side first (which would not _Just Work_ the way Heroku should)
- Using the Pip cache rather than caching our own resources
We should update to rely on pip's caching at some point, but could not find a way to use raw pip to check cache for compatibility with new requirements.txt that also makes sure old dependencies are uninstalled (It may do this for us? But I need more time to spelunk in the source code, and would like to fix this user pain sooner rather than later)


test/run-focus
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
```
schneems: Maybe want to `.gitignore` this? `git rm test/run-focus` `echo test/run-focs >> .gitignore` ?

test/run-deps
```diff
@@ -65,6 +65,16 @@ testPylibmc() {
...
+testDeployingTwiceUsesTheCache() {
+  compile "pythonDefault"
+  assertNotCaptured "Installing collected packages:"
+  assertCapturedSuccess
+  compile "pythonDefault"
+  assertCaptured "Requirement already satisfied:"
+  assertNotCaptured "Installing collected packages:"
+  assertCapturedSuccess
+}
+
```
schneems: Did this end up being useful? I remember it passing even without our patch? We've got the hatchet test as well.

bin/steps/pip-install
```diff
@@ -58,7 +60,12 @@ if [ ! "$SKIP_PIP_INSTALL" ]; then
         exit 1
     fi
 
-    # Smart Requirements handling
+    # Cache pip packages ourselves
+    ## TODO: Update to cache pip's built-in cache
+
+    # # Update cached version of requirements.txt
+    # cp -R $BUILD_DIR/requirements.txt "$CACHE_DIR/.heroku/requirements.txt"
```
schneems: My personal preference is to delete commented out code instead of leaving it. For the TODO, it might be better served in a tracker like the issues.

bin/steps/pip-install
```diff
@@ -47,7 +47,9 @@ if [ ! "$SKIP_PIP_INSTALL" ]; then
     if [ ! -f "$BUILD_DIR/.heroku/python/bin/pip" ]; then
         exit 1
     fi
-    /app/.heroku/python/bin/pip install -r "$BUILD_DIR/requirements.txt" --exists-action=w --src=/app/.heroku/src --disable-pip-version-check --no-cache-dir 2>&1 | tee "$WARNINGS_LOG" | cleanup | indent
+
+
+    /app/.heroku/python/bin/pip install -r "$BUILD_DIR/requirements.txt" --exists-action=w --src=/app/.heroku/src --disable-pip-version-check | tee "$WARNINGS_LOG" | cleanup | indent
```
schneems: It looks like the difference here is removing `--no-cache-dir`. From https://github.com/heroku/heroku-buildpack-python/commit/cc6a03aa541be21edea9ab70bc55a1c4a32bd3d4.

Is this needed for this PR? Can we move it to it's own PR? There's also some extra newlines/whitespace.

bin/compile
```diff
@@ -321,6 +321,8 @@ mtime "sqlite3.install.time" "${start}"
 source "$BIN_DIR/steps/pip-install"
 mtime "pip.install.time" "${start}"
 
+
+
```
schneems: Extra whitespace

bin/compile
```diff
@@ -262,14 +262,14 @@ source "$BIN_DIR/steps/pipenv"
 # Uninstall removed dependencies with Pip.
 # The buildpack will automatically remove any declared dependencies (in requirements.txt)
 # that were explicitly removed. This machinery is a bit complex, but it is not complicated.
-(( start=$(nowms) ))
+# (( start=$(nowms) ))
 # shellcheck source=bin/steps/pip-uninstall
-source "$BIN_DIR/steps/pip-uninstall"
-mtime "pip.uninstall.time" "${start}"
+# source "$BIN_DIR/steps/pip-uninstall"
+# mtime "pip.uninstall.time" "${start}"
```
schneems: Can we clear this?

bin/compile
```diff
@@ -262,14 +262,14 @@ source "$BIN_DIR/steps/pipenv"
...
 
 # If no requirements.txt file given, assume `setup.py develop` is intended.
 # This allows for people to ship a setup.py application to Heroku
-# (which is rare, but I vouch that it should work!)
+# TODO: hatchet test to confirm
```
schneems: I can pair with you on this. My preference would be to move this to an issue though.

CaseyFaist:
Aw heck yes, this is some great feedback @schneems. Updating before deploy, as most recommendations aren't live code.

pip-pop is more than just the pip-diff file we're no longer using - I'd rather do a separate commit to make sure all uses of pip-pop and pip-diff have been found before ripping it out, so leaving it here just for now (and opening an issue)

bin/steps/pip-install
CaseyFaist: good catch, yes this should be in it's own PR

PR #930, opened 2020-03-18
Fix old Django version detection

CaseyFaist:
Running tests from privileged account for #923


bin/steps/pip-install
```diff
@@ -38,10 +38,10 @@ if [ ! "$SKIP_PIP_INSTALL" ]; then
         mcount "failure.none-version"
     fi
 
-    if grep -q 'django==1.*' requirements.txt; then
+    if grep -qi '^django==1.*' requirements.txt; then
```
hone: Does `requirements.txt` accept case insensitive filenames?

danielleadams: I believe this is the "ignore case" flag, right? Is there a reason this needs to be added?

hone: From talking with @CaseyFaist, here's the [PEP](https://www.python.org/dev/peps/pep-0426/#name).

CaseyFaist: pip install is case insensitive 👍

Issue #936, opened 2020-03-26
Cache buildpack version number


CaseyFaist:
An interesting idea from the [ruby buildpack](https://github.com/heroku/heroku-buildpack-ruby) is to cache the release number of the buildpack - this allows Heroku's support team (and me) to zero in on broken changes more quickly.

This would involve some process changes - currently changelog is updated after release, so that changelog isn't pointing to a release that doesn't exist here on Github. Need to chew on this chicken and egg problem a bit more - and open to suggestions 😄

edmorley:
Tracking internally as part of overall observability work.

PR #937, opened 2020-03-26
Update changelog to reflect most recent release

CaseyFaist:
Currently, a release is cut and tagged, and then our changelog is out of date. This is updating after the most recent release, but find discussion for process improvements in #936.


CaseyFaist:
Not today automerge, that ain't right. Superseded by #945

PR #938, opened 2020-03-26
Add Buildpack-Maintainers as code owners

CaseyFaist:
The current owners reflect the other language team members at Heroku, but there are more Herokai interested and qualified to review Prs, whether due to Python experience, bash expertise or other helpful context.

All your reviews are belong to me now!

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,2 @@
 * @heroku/languages
+* @heroku/buildpack-maintainers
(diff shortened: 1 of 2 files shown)


CaseyFaist:
Let's see if requesting your review directly passes the check - I am not sure Changeling will accept your review as fulfilling the compliance check, but if it does, 🎉

edmorley:
I think unfortunately until this PR is merged, the rules from master apply :-)

PR #929, opened 2020-03-18
Codingjoe/shellcheck action

CaseyFaist:
Triggering #900 from privileged account, investigating why changes don't trigger travis test suite

PR #943, opened 2020-03-30
Shellcheck action

codingjoe:
Close #929


edmorley:
@codingjoe Thank you for the PR!

Reading #900 it sounds like one of the motivations for this change was that Travis wasn't working at the time for PRs from forks + to improve parallelism.

Travis now runs on PRs from forks (and everything but the hatchet sub-task passes), and creates many parallel jobs as of #1022. Some of the other buildpacks have also moved from Travis CI to Circle CI, and it's likely the Python buildpack will follow suit in the future.

As such I think it we should leave shellcheck running on Travis for now, and re-evaluate during any future CircleCI move, when I can evaluate the boot time / concurrency limits vs having everything in one place tradeoffs against the long term CI solution rather than Travis.

PR #966, opened 2020-04-24
Prevent Hatchet on forked PRs [changelog skip]

jkutner:
Update travis conditional for hatchet to check for `HEROKU_API_*` env vars


jkutner:
@edmorley i'm doing target practice with a blindfold on right now. Once I get that `if:` condition working we can probably remove the thing in `spec_helper.rb`

PR #973, opened 2020-05-04
Add latest version of setuptools for Python 3.

rvanlaar:
Python2 stays on the original vendered setuptools: 39.0.1.


rvanlaar:
Would need an extra check for setuptools to update it when the current version is lower.

edmorley:
@rvanlaar thank you for the PR! The handling of pip/setuptools changed quite a bit in #1007 which caused conflicts in this PR, and handling also needed to be added for Python 3.4 (since the latest setuptools that supports it is `43.0.0`) - so I ended up fixing this in #1024.

PR #977, opened 2020-05-18
Add support for latest CPython and PyPy versions

edmorley:
Adds support for:
* CPython 2.7.18, 3.5.9, 3.7.7 and 3.8.3
* PyPy2.7 v7.3.1 and PyPy3.6 v7.3.1

The binaries will need generating and uploading before CI will pass.

Note: Whilst the build script for CPython 3.8.3 did already exist in the repository, it appears to have been accidentally created in #920, which predated the existence of that version of Python - so the binaries do not exist on S3.

The Heroku-18 Docker image tag has also been unpinned, since the new libssl version is now available at runtime in all environments, so we don't need to force building against the older version of the headers.

Fixes [W-7582174](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008OnDcIAK/view).


edmorley:
Once I have review, I'll generate/upload the binaries and retrigger CI.

edmorley:
All binaries generated; CI now passes. I'll bump the buildpack version and release next.

PR #1007, opened 2020-07-23
Improve pip/setuptools/wheel installation

### LINKED ISSUE #949: Upgrade setuptools to a more recent version
I'm starting to get incompatibility warnings about the version of setuptools supplied in this buildpack:

>  ERROR: google-auth 1.7.1 has requirement setuptools>=40.3.0, but you'll have setuptools 39.0.1 which is incompatible.

If I manually `pip install` the google-auth package, setuptools is automatically updated but this doesn't seem the case when the app is built.

The version provided in the build pack was released 18 Mar 2018. Latest is 46.1.3

Would it be possible to get an update to setuptools in the next release?

### LINKED ISSUE #999: get-pip.py is downloaded even if it's not used
During the Python setup step, `get-pip.py` is unconditionally downloaded for every build:
https://github.com/heroku/heroku-buildpack-python/blob/156b07ce2b503dd6335708b8d0ea74333b1cece6/bin/steps/python#L139-L148

However it's only used if performing a fresh Python install (eg new app, manually cleared cache or Python/stack version upgrade) or if the Pip version differs:
https://github.com/heroku/heroku-buildpack-python/blob/156b07ce2b503dd6335708b8d0ea74333b1cece6/bin/steps/python#L150-L159

The `get-pip.py` download step should be inside the later conditional, so that the download can be skipped.

Doing this would save time downloading the file and remove another potential failure mode for builds.

### LINKED ISSUE #1000: The setuptools and wheel versions are not fully pinned
As part of Python setup, the buildpack has to manage the installation of pip, setuptools and wheel, so that they are available for the later `pip install` of user-provided dependencies.

Changes in these packages can have a significant impact on the build (particularly since the buildpack has to manually modify site-packages as part of BUILD_DIR -> /app path rewriting, so is sadly sensitive to implementation details), so it's important that we pin to tested versions so that builds don't suddenly break after a new upstream release. 

Equally as important as preventing new releases from being installed unintentionally, is ensuring that:
* these packages in existing apps are upgraded if the buildpack pinned versions change (so they benefit from fixes, and to reduce the version matrix when reasoning about buildpack changes)
* any user-provided custom versions of these packages are cleaned up if they are later removed from `requirements.txt` (to prevent their presence in the build cache causing hard to debug situations)

For pip, the buildpack currently does the right thing:
* if pip is not present, a pinned version is installed (currently `9.0.2` for pipenv users, `19.1.1` for Python 3.4, and `20.0.2` otherwise)
* if pip is present, the version is checked against the expected pinned version and upgraded/downgraded if necessary

However for setuptools, the behaviour is inconsistent:
* for clean installs (new app, manually cleared build cache, Python/stack version upgrade), a pinned setuptools version is installed (currently `39.0.1`)
* for non-clean installs:
  - if the pip version was changed since the last build, then setuptools is upgraded/downgraded to the pinned version (if necessary)
  - otherwise, the setuptools version (or even the presence of setupools at all) is not checked/managed

And for wheel, the situation is worse:
* for clean installs, the wheel version is not pinned so the latest is installed
* for non-clean installs, so long as any version of wheel is installed, it's never upgraded/changed

The above leads to both unpredictability of installed versions, as well as confusing UX when performing other actions (eg failures after upgrading to a new Python patch version, that had nothing to do with the Python upgrade itself, but just happened to trigger installing newer setuptools/wheel etc).

We should pin all three to specific versions, and ensure that these versions are checked/managed during each build, and not just after a cache-clearing event.

See:
https://github.com/heroku/heroku-buildpack-python/blob/156b07ce2b503dd6335708b8d0ea74333b1cece6/bin/steps/python#L150-L161

### LINKED ISSUE #1001: Setuptools is first installed from PyPI then installed again using the vendored version
The Python setup step uses get-pip.py to install pip, here:
https://github.com/heroku/heroku-buildpack-python/blob/156b07ce2b503dd6335708b8d0ea74333b1cece6/bin/steps/python#L159

Afterwards the vendored setuptools wheel is installed, here:
https://github.com/heroku/heroku-buildpack-python/blob/156b07ce2b503dd6335708b8d0ea74333b1cece6/bin/steps/python#L160

However `get-pip.py` actually installs setuptools itself (if it's not already installed):
https://pip.pypa.io/en/stable/installing/#installing-with-get-pip-py
https://github.com/pypa/get-pip/blob/eff16c878c7fd6b688b9b4c4267695cf1a0bf01b/templates/default.py#L152-L153

This means the latest version of setuptools (currently `49.2.0`) is installed from PyPI, and then immediately after the vendored copy (currently `39.0.1`) installed over it.

This adds time to the build unnecessarily, and negates any reliability/security benefits of vendoring setuptools.

We should either:
1. Disable the automatic get-pip.py setuptools installation using `--no-setuptools`, and install only the vendored version
2. Stop vendoring setuptools and instead pin the setuptools version by passing `setuptools==VERSION` to get-pip, so the copy it downloads from PyPI is the version we want rather than latest

### LINKED ISSUE #1002: Errors during the install of pip/setuptools/wheel don't cause the build to fail
Whilst much of the buildpack is run with bash's exit on error enabled, the python setup step disables that mode temporarily using `set +e`:
https://github.com/heroku/heroku-buildpack-python/blob/156b07ce2b503dd6335708b8d0ea74333b1cece6/bin/steps/python#L3

Most of the logic in that file is written to take this into account, with the exception of the pip/setuptools/wheel install here:
https://github.com/heroku/heroku-buildpack-python/blob/156b07ce2b503dd6335708b8d0ea74333b1cece6/bin/steps/python#L159-L160

If either of those commands fail (eg PyPI down, transient network issue, pip/buildpack bug), then the build continues regardless, leaving the Python install without a copy of pip/setuptools/wheel - which will cause confusing failures in later buildpack steps.

To compound the issue, the stderr from those commands is sent to `/dev/null`, making it seem like the step succeeded / obscuring the reason for the failed install.

We should:
1. Re-enable exit-on-error mode for these steps
2. Either redirect only stdout to `/dev/null`, or remove all redirection and use the `--quiet` options amongst others to suppress non-error output

### LINKED ISSUE #1003: The installed pip version check is susceptible to false negatives
As part of the Python step step, the pip version is checked here:
https://github.com/heroku/heroku-buildpack-python/blob/156b07ce2b503dd6335708b8d0ea74333b1cece6/bin/steps/python#L151

`pip --version` outputs stdout of form:
```
$ pip --version
pip 20.1.1 from /usr/local/lib/python3.8/site-packages/pip (python 3.8)
```

And `$PIP_UPDATE` is set to a pip version string such as `20.0.2`:
https://github.com/heroku/heroku-buildpack-python/blob/156b07ce2b503dd6335708b8d0ea74333b1cece6/bin/compile#L67

The way the version check conditional is currently written means that the check will succeed if `$PIP_UPDATE` happens to be a substring of the real pip version, even if the version is not an exact match.

If pip always used a versioning scheme of `X.Y.Z` this would not be a problem, however for new minor version releases pip uses versions of form `X.Y` instead of `X.Y.0` (ie: `20.1` instead of `20.1.0`). This means that some release versions are a substring of another version (`20.1` is a substring of `20.1.1`).

As such, if we were ever to downgrade `PIP_UPDATE` from say `20.1.1.` to `20.1` in response to a regression (or if an app rolled back to a previous buildpack release that used the older version), then the pip version would be seen as matching, and not trigger a downgrade as expected.

We should either:
1. Use something like `*" $PIP_UPDATE "*` (with spaces padding) instead of `*$PIP_UPDATE*`
2. Switch to a way of fetching the pip version that doesn't require matching substrings (eg `python -c 'from pip import __version__; print(__version__)'`)
3. Let pip decide whether it's up to date itself, by just unconditionally calling get-pip.py with the target pip version

edmorley:
Notable changes:

- The installed versions of setuptools/wheel are now deterministic
- The build output now includes the versions of pip/setuptools/wheel used, making it easier to debug future upgrades
- Errors during pip/setuptools/wheel install now correctly fail the build, and error output is no longer sent to `/dev/null`
- Pip is now installed using itself instead of `get-pip.py`
- Setuptools is installed from PyPI rather than a vendored copy
- Various other small fixes

**For more details (and easier review) see the individual commit messages and diffs.**

Fixes #1000.
Fixes #1001.
Fixes #1002.
Fixes #1003.
Closes #939.
Closes #999.


bin/steps/python
```diff
@@ -131,34 +131,48 @@ if [ ! "$SKIP_INSTALL" ]; then
...
-if ! curl -s "${GETPIP_URL}" -o "$GETPIP_PY" &> /dev/null; then
-    mcount "failure.python.get-pip"
-    echo "Failed to pull down get-pip"
-    exit 1
+PIP_VERSION='20.1.1'
+# Must use setuptools <47.2.0 until we fix: https://github.com/heroku/heroku-buildpack-python/issues/1006
+SETUPTOOLS_VERSION='47.1.1'
+WHEEL_VERSION='0.34.2'
+
+if [[ "${PYTHON_VERSION}" == ${PY34}* ]]; then
```
dzuelke, line 145: What happens for Python 3 older than 3.4?

bin/steps/python
```diff
@@ -131,34 +131,48 @@ if [ ! "$SKIP_INSTALL" ]; then
...
+#    the install, since get-pip.py can't install the embedded version directly)
+#  - we would still have to manage several versions of get-pip.py, to support older Pythons.
+# Instead, we use the pip wheel to install itself, using the method described here:
+# https://github.com/pypa/pip/issues/2351#issuecomment-69994524
 
-  # Remove old installations.
-  rm -fr /app/.heroku/python/lib/python*/site-packages/pip-*
-  rm -fr /app/.heroku/python/lib/python*/site-packages/setuptools-*
+PIP_WHEEL_FILENAME="pip-${PIP_VERSION}-py2.py3-none-any.whl"
+PIP_WHEEL_URL="https://lang-python.s3.amazonaws.com/common/${PIP_WHEEL_FILENAME}"
```
dzuelke, line 166: Should this be using `$VENDOR_URL` rather than hardcoding the entire thing?

dzuelke:
I'd suggest two things:

1. break this up into more than one commit (and more than one changelog line) if at all possible
2. tests? ;)

bin/steps/python
edmorley: It's already broken (Pip 18.0 dropped support: https://pip.pypa.io/en/stable/news/#id310), and binaries for 3.3 only exist on Cedar-14.

bin/steps/python
edmorley: `$VENDOR_URL` includes the `$STACK`, whereas these are not stack specific.

Refactoring the URL handling to separate out S3 base URL is on my list, but given the existing get-pip.py URL this PR replaces also hardcodes the URL, this seemed like something to do separately.

edmorley: Some context I forgot to mention:
* Currently `VENDOR_URL` is leaked to subprocesses such as `bin/post_compile`, so changing it has the chance of being a breaking change
* Adding an `S3_BASE_URL` and then using `VENDOR_URL="${S3_BASE_URL}/${STACK}` etc means that this feature would also really need rewriting/removing/...: https://github.com/heroku/heroku-buildpack-python/blob/d9b1c73f637f136eac038f526d161d83b94168d8/bin/compile#L43-L48

At the moment there are so many things to fix, many of which interact with each other, that I'm trying to avoid the temptation to scope creep each change too far.

edmorley:
I've:
* moved the pip/setuptools version upgrades out of this PR
* split up the commits
* added some tests

PR #1011, opened 2020-07-24
Reduce the number of env vars exposed to subprocess

### LINKED ISSUE #1010: Several buildpack script variables are leaked into the environment of subprocesses
The environment variables set by the build system are of form:

```
BUILDPACK_LOG_FILE=/dev/fd/3
DYNO=run.4869
HOME=/app
LANG=en_US.UTF-8
PATH=/usr/local/bin:/usr/bin:/bin
PWD=/tmp/codon/tmp/buildpacks/61a71cfe58e0f4e33ca0dffe72ee3f58ebe50e3b
REQUEST_ID=f665634a-3d90-bd3e-3b85-1da2d62c25c6
SHLVL=2
SOURCE_VERSION=827a7d6cdb8f3131b69c14940ecc36811eabf1aa
STACK=heroku-18
_=/usr/bin/env
```
(Generated using a custom buildpack that outputs `env | sort`)

When the Python buildpack runs `bin/post_compile`, the environment is the above plus:

```
BIN_DIR=/tmp/codon/tmp/buildpacks/0f40890b54a617ec2334fac0439a123c6a0c1136/bin
BPLOG_PREFIX=buildpack.python
BUILD_DIR=/tmp/build_1614145f
CACHE_DIR=/tmp/codon/tmp/cache
CACHED_PYTHON_STACK=heroku-18
C_INCLUDE_PATH=/app/.heroku/vendor/include:/app/.heroku/python/include:
CPLUS_INCLUDE_PATH=/app/.heroku/vendor/include:/app/.heroku/python/include:
DEFAULT_PYTHON_STACK=cedar-14
DEFAULT_PYTHON_VERSION=python-3.6.11
ENV_DIR=/tmp/d20200724-66-1g7dm3a
EXPORT_PATH=/tmp/codon/tmp/buildpacks/0f40890b54a617ec2334fac0439a123c6a0c1136/bin/../export
LATEST_27=python-2.7.18
LATEST_34=python-3.4.10
LATEST_35=python-3.5.9
LATEST_36=python-3.6.11
LATEST_37=python-3.7.8
LATEST_38=python-3.8.5
LD_LIBRARY_PATH=/app/.heroku/vendor/lib:/app/.heroku/python/lib:
LIBRARY_PATH=/app/.heroku/vendor/lib:/app/.heroku/python/lib:
PIP_UPDATE=20.0.2
PKG_CONFIG_PATH=/app/.heroku/vendor/lib/pkgconfig:/app/.heroku/vendor/lib/pkg-config:/app/.heroku/python/lib/pkg-config:
PROFILE_PATH=/tmp/build_1614145f/.profile.d/python.sh
PY27=python-2.7
PY34=python-3.4
PY35=python-3.5
PY36=python-3.6
PY37=python-3.7
PYPY_27=pypy2.7-7.3.1
PYPY_36=pypy3.6-7.3.1
PYTHONUNBUFFERED=1
RECOMMENDED_PYTHON_VERSION=python-3.6.11
VENDOR_URL=https://lang-python.s3.amazonaws.com/heroku-18
WARNINGS_LOG=/tmp/tmp.ST9FBfCWKz
```

Of those, several are internal variables that shouldn't be exposed:

```
BPLOG_PREFIX=buildpack.python
CACHED_PYTHON_STACK=heroku-18
DEFAULT_PYTHON_STACK=cedar-14
DEFAULT_PYTHON_VERSION=python-3.6.11
LATEST_27=python-2.7.18
LATEST_34=python-3.4.10
LATEST_35=python-3.5.9
LATEST_36=python-3.6.11
LATEST_37=python-3.7.8
LATEST_38=python-3.8.5
PIP_UPDATE=20.0.2
PY27=python-2.7
PY34=python-3.4
PY35=python-3.5
PY36=python-3.6
PY37=python-3.7
PYPY_27=pypy2.7-7.3.1
PYPY_36=pypy3.6-7.3.1
RECOMMENDED_PYTHON_VERSION=python-3.6.11
WARNINGS_LOG=/tmp/tmp.ST9FBfCWKz
```
(I've excluded a few that are technically internal, but have a higher chance of already being relied upon, for example `$PROFILE_PATH`)

The reason this is occurring, is that `export` is used fairly liberally throughout the buildpack for variables, even though most scripts are sourced (so variables can still be accessed without `export`).

edmorley:
The following env vars are no longer exposed to subprocesses run by the buildpack (such as the `bin/pre_compile` and `bin/post_compile` hooks):

* `BPLOG_PREFIX`
* `CACHED_PYTHON_STACK`
* `DEFAULT_PYTHON_STACK`
* `DEFAULT_PYTHON_VERSION`
* `LATEST_27`
* `LATEST_34`
* `LATEST_35`
* `LATEST_36`
* `LATEST_37`
* `LATEST_38`
* `PIP_UPDATE`
* `PY27`
* `PY34`
* `PY35`
* `PY36`
* `PY37`
* `PYPY_27`
* `PYPY_36`
* `RECOMMENDED_PYTHON_VERSION`
* `WARNINGS_LOG`

There were previously no tests at all for the pre/post-compile hooks, so I've added some now.

Fixes #1010.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 # Master
 
+- Reduce the number of environment variables exposed to `bin/{pre,post}_compile` and other subprocesses (#1011)
 
 # 173 (2020-07-21)
 
--- a/bin/compile
+++ b/bin/compile
@@ -16,7 +16,9 @@
 set -eo pipefail
 
 # Boostrap the Buildpack Standard Library.
-export BPLOG_PREFIX="buildpack.python"
+# Disable unused env var warning since shellcheck doesn't know about the stdlib.
+# shellcheck disable=2034
+BPLOG_PREFIX="buildpack.python"
 export BUILDPACK_LOG_FILE=${BUILDPACK_LOG_FILE:-/dev/null}
 
 [ "$BUILDPACK_XTRACE" ] && set -o xtrace
@@ -84,16 +86,12 @@ if [[ -f "$BUILD_DIR/Pipfile" ]]; then
     PIP_UPDATE="9.0.2"
 fi
 
-export DEFAULT_PYTHON_STACK PIP_UPDATE
-export PY37 PY36 PY35 PY27 PY34
-
 # Common Problem Warnings:
 # This section creates a temporary file in which to stick the output of `pip install`.
 # The `warnings` subscript then greps through this for common problems and guides
 # the user towards resolution of known issues.
 WARNINGS_LOG=$(mktemp)
-export WARNINGS_LOG
-export RECOMMENDED_PYTHON_VERSION=$DEFAULT_PYTHON_VERSION
+RECOMMENDED_PYTHON_VERSION=$DEFAULT_PYTHON_VERSION
 
 # The buildpack ships with a few executable tools (e.g. pip-grep, etc).
 # This installs them into the path, so we can execute them directly.
@@ -216,8 +214,6 @@ if [ -f "$CACHE_DIR/.heroku/python-stack" ]; then
 else
   CACHED_PYTHON_STACK=$STACK
 fi
-
-export CACHED_PYTHON_STACK
 
 # Pipenv Python version support.
 # Detect the version of Python requested from a Pipfile (e.g. python_version or python_full_version).
--- a/bin/default_pythons
+++ b/bin/default_pythons
@@ -1,5 +1,10 @@
 #!/usr/bin/env bash
 
+# Disable unused env var warning, since shellcheck doesn't take into account
+# that this file is sourced. We don't want to use export since it exposes
+# the env vars to subprocesses.
+# shellcheck disable=2034
+
 DEFAULT_PYTHON_VERSION="python-3.6.11"
 LATEST_38="python-3.8.5"
 LATEST_37="python-3.7.8"
@@ -9,13 +14,3 @@ LATEST_34="python-3.4.10"
 LATEST_27="python-2.7.18"
 PYPY_36="pypy3.6-7.3.1"
 PYPY_27="pypy2.7-7.3.1"
-
-export DEFAULT_PYTHON_VERSION \
-    LATEST_38 \
-    LATEST_37 \
-    LATEST_36 \
-    LATEST_35 \
-    LATEST_34 \
-    LATEST_27 \
-    PYPY_36 \
-    PYPY_27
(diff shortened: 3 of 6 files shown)


edmorley:
@Malax I've updated the PR to fix the test runner (by adding `run_in_clean_env`) to make the tests pass. Previously env vars from outside the test runner were leaking into the tests. This was an issue since the way tests run locally and on Travis differs even though they both use Docker (something else to fix at some point) so the test was reporting different env vars for each.

edmorley:
> I've updated the PR to fix the test runner

This fix required several broken tests to be fixed too, so I split those changes out to #1016, and have now rebased on that.

edmorley:
Even after the test harness fixes in #1016 the newly added test here was failing -- and seemingly only on Travis (it passed locally). 

After two hours of digging into the differences between the way the tests run locally and on Travis (something that's now resolved in #1021), I noticed that actually the failures seen were not Travis-specific but stack-specific, and locally I happened to be using the one stack out of three on which the test passed (Heroku-18) 🤦 

The fact that the tests failed on only 2 out of 3 stacks on Travis was hard to spot, since the test matrix was configured to split jobs by test suite and not stack, and Heroku-18 was the last stack to run in each job so the passing test was hidden right at the end of the log. I've now split the stacks out into their own jobs as part of #1018 to prevent this in the future.

The test failure itself was due to `OLDPWD` being present in the environment of `bin/{pre,post}_compile` on Heroku-18, but not for Cedar-14 or Heroku-16.

I managed to create a repro outside the tests:

Ubuntu 16.04:
```
$ docker run --rm -it ubuntu:16.04 bash -c 'cd /tmp; echo "OLDPWD=$OLDPWD"; bash -c "echo \"OLDPWD=\$OLDPWD\""'
OLDPWD=/
OLDPWD=
```

Ubuntu 18.04:
```
$ docker run --rm -it ubuntu:18.04 bash -c 'cd /tmp; echo "OLDPWD=$OLDPWD"; bash -c "echo \"OLDPWD=\$OLDPWD\""'
OLDPWD=/
OLDPWD=/
```

The version of bash in Ubutnu 16.04 is `4.3.48`, and in 18.04 is `4.4.20`.

As such I believe it's due to this change:
`OLDPWD can be inherited from the environment if it names a directory.`
(https://git.savannah.gnu.org/cgit/bash.git/tree/NEWS)

Which is referenced also here:
https://unix.stackexchange.com/questions/242909/why-does-bash-clear-oldpwd-when-a-child-script-is-started
https://groups.google.com/d/msg/gnu.bash.bug/k7o0xnVyUzI/MJz9SfMXAQAJ

It seems fine to just exclude that env var from the test.

PR #1022, opened 2020-07-28
Travis: Reduce end to end testing time

### LINKED ISSUE #1018: Speed up the Travis run
Currently the Travis run takes ~22 mins wall time (45-50 mins job time), which gets pretty tedious when performing tasks that involve several PRs in sequence (such as releasing a new Python version and then buildpack release).

Longer term we will likely be moving more of the end to end tests to Hatchet, and using bash tests for just unit testing. The Hachet tests can then be parallelised in a similar manner to what the Ruby buildpack and others use already.

However in the meantime we should:

- adjust (or reduce) the use of Travis job stages
- consider splitting up the unit tests by stack (currently each entry in the matrix builds multiple stacks, wasting image pulling time)
- use the correct docker images (see #958), which saves having to spend additional time building the wrong ones
- only installing the ruby deps for the hatchet job

### LINKED ISSUE #1019: Travis tries to run Hatchet tests on PRs from forks
The Hatchet tests run against the platform, so require Heroku credentials. These are specified in the form of `HEROKU_API_USER` and `HEROKU_API_KEY` environment variables set on the buildpack's Travis config. By design Travis does not make environment variables marked as secure available to travis runs from PRs on forks (otherwise the values could be printed/...).

As such the Travis config currently tries to skip the Hatchet tests for such PRs, to prevent the tests from failing:

https://github.com/heroku/heroku-buildpack-python/blob/f21e538fde101f3dd01f91c41db64963a6694fb3/.travis.yml#L27

Unfortunately this doesn't seem to work, since the tests are still being run and fail with an error message that may not be clear for contributors:
https://travis-ci.com/github/heroku/heroku-buildpack-python/jobs/362304689#L309

We should adjust the conditional to fix this - perhaps by checking for the presence of `HEROKU_API_USER` and `HEROKU_API_KEY` directly.

edmorley:
This change (along with #1021, which skips an unnecessary docker build) reduces the wall clock time from ~22 minutes to ~6 minutes. Even with the additional Travis setup overhead from the larger job matrix, the combined job duration (~50 minutes) has not increased due to the other time savings.

Changes:

- for the unit tests, each stack is now tested in its own job and so tested in parallel
- the use of Travis stages has been removed, since by design it blocks later tasks on earlier stages having completed - reducing parallelism unnecessarily for this use case
- all jobs except for the Hatchet job now use Travis' `minimal` image, and no longer install redundant Ruby + bundler
- the `sudo: {required,false}` references have been removed, since Travis no longer supports its non-sudo container infrastructure so ignores that option

Fixes #1018.

[skip changelog]

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,37 +1,40 @@
-language: ruby
+language: minimal
 dist: bionic
-sudo: required
 branches:
   only:
   - master
 
-rvm:
-- 2.6.6
-before_script:
-  - gem install bundler -v 1.16.2
-
 script:
-  - make test STACK="cedar-14" TEST_CMD="${TESTFOLDER}"
-  - make test STACK="heroku-16" TEST_CMD="${TESTFOLDER}"
-  - make test STACK="heroku-18" TEST_CMD="${TESTFOLDER}"
+  - make test STACK="${STACK}" TEST_CMD="${TEST_CMD}"
 
 jobs:
   include:
-  - stage: Bash linting (shellcheck)
-    sudo: false
+  - name: Bash linting (shellcheck)
     script: make check
-  - stage: Hatchet Integration
+  - name: Hatchet integration tests
     if: env(TRAVIS_PULL_REQUEST_SLUG) = env(TRAVIS_REPO_SLUG)
-    name: Run Hatchet
+    language: ruby
+    rvm:
+      - 2.6.6
+    before_script:
+      - gem install bundler -v 1.16.2
     script:
       - bundle exec hatchet ci:setup
       - PARALLEL_SPLIT_TEST_PROCESSES=11 bundle exec parallel_split_test spec/hatchet/
 
 env:
-  matrix:
-    - TESTFOLDER=test/run-deps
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,37 +1,40 @@
...
     script: make check
-  - stage: Hatchet Integration
+  - name: Hatchet integration tests
     if: env(TRAVIS_PULL_REQUEST_SLUG) = env(TRAVIS_REPO_SLUG)
-    name: Run Hatchet
+    language: ruby
+    rvm:
+      - 2.6.6
+    before_script:
+      - gem install bundler -v 1.16.2
```
danielleadams, line 20: Is there a reason for bundler 1.16.x? The last 1.x is 1.17.3

edmorley: Yeah I wondered that too - I left as is since I was planning on updating all the hatchet-related dependencies at some point, and this version is pre-existing.

edmorley: Updated in #1036

Issue #1028, opened 2020-08-03
Switch default branch from 'master' to 'main'


edmorley:
As part of ensuring inclusive language is used throughout all Heroku repositories, we should switch from the current default branch of `master` to `main`.

This will involve:

- [x] Creating a new branch `main`, based off current `master`
- [x] Changing the default branch in GitHub to `main`
- [x] Updating the branch protection settings
- [x] Updating the target branch of any open PRs
- [x] Updating any in-repo code/docs references (such as the Travis badge) -> #1029
- [x] Checking that there are no branch-specific Travis environment variables set
- [x] Updating the weekly Travis cron to build from the new branch
- [x] Checking that Travis still works -> https://travis-ci.com/github/heroku/heroku-buildpack-python/builds/178287518
- [x] Checking for any other deep links to this repository in other repos that include the branch name
- [x] ~Creating a support briefing to mention the change~ Email the affected app owners
- [x] Deleting the `master` branch

Note:
Most users will be consuming this buildpack via the buildpack-registry stable release (using the shorthand alias `heroku/python`). Some will be using the master branch via a buildpack URL of `https://github.com/heroku/heroku-buildpack-python`, and others will be pinned to a specific branch or tag using `https://github.com/heroku/heroku-buildpack-python#some-branch-or-tag`.

In rare cases people may be using the latter form to specify the master branch, even though it's redundant to specify the branch name. Such usages will need to be updated to use either the default-branch GitHub form, or else the shorthand form (recommended).

Also GitHub now redirects any web links (such as the `/blob/` URLs) whose branch doesn't exist, which reduces the urgency of catching every stray docs reference:
https://github.blog/changelog/2020-07-17-links-to-deleted-branches-now-redirect-to-the-default-branch/

@W-7837248@

edmorley:
> In rare cases people may be using the latter form to specify the master branch

Analysing buildpack URLs used, I found only 2 apps using `#master` with this buildpack, so I will contact the app owners directly prior to deleting the old `master` branch.

PR #1029, opened 2020-08-03
Replace 'master' branch references with 'main'

edmorley:
For #1028.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: minimal
 dist: bionic
 branches:
   only:
-  - master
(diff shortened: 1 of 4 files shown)


edmorley:
@amureki Hi - thank you for reviewing!

I spotted that reference, though it's in a test fixture which is a vendored version of https://github.com/heroku/python-getting-started -- so I was going to wait until that repo was updated before potentially re-vendoring in a new version (though we may end up replacing the unit test with a hatchet test instead, in which case we would just remove the vendored files).

PR #1033, opened 2020-08-04
Symlink staticfiles option

jordaneb:
Django's `collectstatic` command has a `--link` option (https://docs.djangoproject.com/en/2.2/ref/contrib/staticfiles/#cmdoption-collectstatic-link) which symlinks files where possible instead of creating copies of them.

I've seen this reduce the size of `collected-static` directories by up to 40% so this should have a positive impact on slug sizes!

As far as I can tell this is also compatible with Whitenoise and doesn't interfere with how it generates gzipped versions of files as those remain as actual files.

### DIFF
--- a/bin/steps/collectstatic
+++ b/bin/steps/collectstatic
@@ -9,6 +9,7 @@
 # Runtime arguments:
 #   - $DISABLE_COLLECTSTATIC: disables this functionality.
 #   - $DEBUG_COLLECTSTATIC: upon failure, print out environment variables.
+#   - $SYMLINK_COLLECTSTATIC: symlink static files where possible.
 
 # shellcheck source=bin/utils
 source "$BIN_DIR/utils"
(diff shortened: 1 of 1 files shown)


edmorley (changes requested):
Hi! Thank you for the PR :-)

This definitely sounds like something that could be a win (both in terms of build time and slug size) and so something we should do.

My initial thoughts/questions are:
* what versions of Django support `--link` and are there any known bugs that prevent its use on older versions?
* are there any scenarios where it's not a drop in replacement (eg alternative staticfiles backends, or ...?)
* should we make this opt out rather than opt in? (either now or eventually?)
* we should probably ensure the metrics emitted allow determining pass vs fail and duration for link vs standard
* we'll need tests for this

edmorley:
I've filed a corresponding issue for this feature request as #1060 (in part so I can sync to our internal tracking tools, so I can ensure it's in the backlog).

edmorley:
> I've filed a corresponding issue for this feature request as #1060 (in part so I can sync to our internal tracking tools, so I can ensure it's in the backlog).

Closing this out for now - let's track resolving the questions mentioned above in #1060 :-)

PR #1059, opened 2020-08-27
use USE_STAGING_BINARIES splicing PIP_WHEEL_URL

duanhongyi:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,12 @@
 # Python Buildpack Changelog
 
 ## Unreleased
+- Custom PIP_WHEEL_URL environment variable (#1058).
 
 
 ## v177 (2020-08-18)
 
-- Python 3.6.12 and 3.7.9 are now available (CPython) (#1054).
+- Python 3.6.12 and 3.7.9 are now available (CPython) (#1059).
 - The default Python version for new apps is now 3.6.12 (previously 3.6.11) (#1054).
 
 ## v176 (2020-08-12)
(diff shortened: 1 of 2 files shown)


edmorley (changes requested):
Hi! Thank you for the PR :-)

Could you describe the motivation for this change? (For example is what's the use-case and is this for Heroku or something else?) Generally a PR should always have a description that describes the "why" that isn't necessarily apparent from the "what" of the diff itself. There's a good guide to commit messages here that may be of interest :-)
https://chris.beams.io/posts/git-commit/

If this is something that is accepted, it will also:
* need some tests
* need to use a different approach for the env var, depending on whether this is for Heroku use or not (since app env vars aren't automatically set in the build enviroment; buildpacks have to load ENV_DIR etc)

Many thanks!

edmorley:
Oops forgot to also say - if you make further changes to the PR, please add them to this one - no need to make a new PR each time :-)

edmorley:
@duanhongyi Hi! I don't suppose you saw my questions above? :-)

duanhongyi:
I'm really sorry. I just came back from my vacation.

The motivation for this RP is simple. In my opinion, in `USE_STAGING_BINARIES` should contain the PIP installation script.

edmorley:
> The motivation for this RP is simple. In my opinion, in USE_STAGING_BINARIES should contain the PIP installation script.

@duanhongyi Hi! Thank you for the reply. However I'm still not sure what the motivation is? For example:
* what are you trying to do with the buildpack
* why does the current implementation make that hard
* what are you doing to work around it in the meantime
* how does this change make that easier?

Without more details we won't be able to accept the change.

edmorley:
@duanhongyi Hi! I suspect your use-case might be fixed by #1085 (which I needed to do for other reasons, but conveniently makes custom S3 URLs for the pip case easier too).

PR #1076, opened 2020-09-22
fix small typo re: install of python packages from git

ckcollab:
simple typo fix, found when I was checking into how caching works!


edmorley (approved):
Thank you for the PR!

(I've triggered CI on a non-fork PR #1077 (that has access to the credentials required for the Hatchet CI step to pass) since the compliance check requires that CI run even for comment changes like this.)

PR #1087, opened 2020-10-01
Tests: Make assertCapturedSuccess check stderr is empty again

### LINKED ISSUE #1025: assertCapturedSuccess no longer asserts that STDERR is empty
The `assertCapturedSuccess` assertion in `test/utils` used to check both the exit code was zero, and that stderr was empty.

However in 797652a75d69a1fece96a26bf4514fe9e9e1c020 the stderr part of the assertion was commented out without explanation:
https://github.com/heroku/heroku-buildpack-python/commit/797652a75d69a1fece96a26bf4514fe9e9e1c020#diff-65c067a6f0a3aef292fb54ec21a1fe8cR98

This also has the side effect of meaning non-empty stderr is never shown for non-zero exit codes, due to this later conditional:
https://github.com/heroku/heroku-buildpack-python/blob/f21e538fde101f3dd01f91c41db64963a6694fb3/test/utils#L90

...which makes debugging harder.

I'm presuming the reason was that there may be some cases where we are fine with a warning on stderr so long as the exit code is zero. However we should still validate stderr where possible.

edmorley:
`assertCapturedSuccess` used to check that `stderr` was empty, until:
https://github.com/heroku/heroku-buildpack-python/commit/797652a75d69a1fece96a26bf4514fe9e9e1c020#diff-65c067a6f0a3aef292fb54ec21a1fe8cR98

Adding back the assertion exposes some new bugs, which I'll fix in later PRs. In the meantime affected tests have been adjusted to use a new `assertCapturedSuccessWithStdErr`.

The test harness output for the failing case has also been improved, to ease debugging.

Closes [W-7918492](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008WooEIAS/view).
Closes #1025.

[skip changelog]

### DIFF
--- a/test/run-deps
+++ b/test/run-deps
@@ -43,7 +43,9 @@ testNLTK() {
   echo 'ignore::RuntimeWarning' > "${env_dir}/PYTHONWARNINGS"
   compile 'nltk' '' "${env_dir}"
   assertCaptured "[nltk_data] Downloading package city_database" "STD_ERR"
-  assertCapturedSuccess
+  # Can't use `assertCapturedSuccess` since the NLTK downloader outputs all
+  # progress/status messages to stderr (W-8146040).
+  assertCapturedSuccessWithStdErr
 }
 
 testPsycopg2() {
--- a/test/run-features
+++ b/test/run-features
@@ -33,7 +33,9 @@ testStackChange() {
 testSetupPy() {
   compile "setup-py"
   assertCaptured "maya"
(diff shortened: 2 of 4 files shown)


edmorley:
(This change means the tests now catch a few failure cases which would otherwise have been silently introduced once #968 is updated/merged)

danielleadams (approved):

PR #1092, opened 2020-10-07
Tests: Clean up the Python version unit tests

edmorley:
* Fixes the `Installing <version>` assertions so that they don't false positive against the ~`please upgrade to <version>` output.
* Removes modification of test fixtures during tests, since it can lead to failures depending on test order, and confusion when debugging.
* Updates the PyPy version warning tests to use a slightly newer (but still not latest) PyPy version, which means that the test now passes on Cedar-14 and can be unskipped.
* Switches to using an empty requirements file for version tests that duplicate the main test, to save spending time installing dependencies unnecessarily.
* Switches the NLTK test to using the default buildpack Python version, rather than an ancient Python 3.6.
* Skips the Python 3.7/3.8 tests on Cedar-14 rather than asserting failure, since we know they'll never pass due to Cedar-14's libssl being older than required.
* Removes redundant `testSqliteInstall` test since it duplicates the Python version install tests.

Longer term I'll be moving many of the unit tests to Hatchet, however this at least makes the tests more dependable in the meantime.

Closes [W-8060219](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008cuSbIAI/view).
Closes [W-8176779](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008gUoAIAU/view).

[skip changelog]


danielleadams (approved):

test/run-versions
```diff
@@ -4,59 +4,58 @@
...
   # and the newest Pip that works on Python 3.4 doesn't support `PIP_NO_PYTHON_VERSION_WARNING`.
   assertCapturedSuccessWithStdErr
 }
 
 testPython3_4_warn() {
   compile "python3_4_warn"
-  assertCaptured "python-3.4.9"
+  assertCaptured "Installing python-3.4.9"
   assertCaptured "security update!"
+  assertCaptured "${LATEST_34}"
```
vanastassiou, line 50: Can I double-check with you about this and the corresponding test for 3.5? It doesn't make sense to me to include both a check for the `security update!` string and the most recent minor version. Running `make test` locally complains at this line and the test for 3.5 at line 58.

edmorley: @vanastassiou Hi! In the case of a security update being available, the warning message shown says:

> Python has released a security update! Please consider upgrading to `<latest version>`

This is what that assertion is checking against.

See:
https://github.com/heroku/heroku-buildpack-python/blob/fcf696b835d7abce1631d841de7ae5fdd5dde45c/bin/steps/python#L46-L59

The tests for this repo are passing on `main`:
https://travis-ci.com/github/heroku/heroku-buildpack-python

Are you using an older branch/fork, or else have set a custom `BUILDPACK_S3_BASE_URL` pointing at a bucket that's missing those files?

test/run-versions
```diff
@@ -65,15 +64,15 @@ testPython3_5() {
 
 testPython3_5_warn() {
   compile "python3_5_warn"
-  assertCaptured "python-3.5.6"
+  assertCaptured "Installing python-3.5.6"
   assertCaptured "security update!"
```
vanastassiou, line 68: I don't think the "security update" string appears anymore when installing an outdated version of 3.5, either.

edmorley: @vanastassiou The tests are passing on Travis, and this works for me in production:

```
$ git init
$ h create edmorley---buildpack https://github.com/heroku/heroku-buildpack-python
$ echo 'python-3.5.9' > runtime.txt
$ touch requirements.txt
$ git add -A && git commit -m '.' && git push heroku main
...
remote: -----> Python app detected
remote:  !     Python has released a security update! Please consider upgrading to python-3.5.10
remote:        Learn More: https://devcenter.heroku.com/articles/python-runtimes
remote: -----> Installing python-3.5.9
remote: -----> Installing pip 20.1.1, setuptools 47.1.1 and wheel 0.34.2
```

If you are running the tests for this repository locally, could you try cloning https://github.com/heroku/heroku-buildpack-python to a new directory to rule out modifications to the existing clone?

Or if you are seeing this in production, please could you file a support ticket at https://help.heroku.com

Issue #1093, opened 2020-10-08
Allow to override pip version for pipenv users


vityas:
## question
Assuming that we don't want to build a new buildpack, is there a workaround for pipenv users to be able to install packages that require a newer pip? unpin and override pip.

## reason
A common example of a package that will fail to install is `tensorflow>=2.0`

## side question
I doubt any developers keep using `pip==9.0.2`. It seems like the main and only reason to pin 9.0.2 is that (https://github.com/heroku/heroku-buildpack-python/issues/786) 

it "would require users to uninstall and reinstall their local pipenv environments"

 Not sure how some one-time inconvenience can outweigh the need to have the same environment. which simply not possible with the old pip.

## example PR
see the attached PR. I created it as an example. It might not be the best solution, I would love to see some workarounds.

edmorley:
Hi! I've left a comment on the PR:
https://github.com/heroku/heroku-buildpack-python/pull/1094#pullrequestreview-506483377

timmygee:
Just flagging that the legend @alan-eu has created a fork that does exactly this

https://github.com/alan-eu/heroku-buildpack-python

edmorley:
Hi! The pipenv version has just been updated in #1169.

I'm closing this out since adding env var support for pipenv/pip is something that I believe will introduce more issues than it fixes for the reasons mentioned in:
https://github.com/heroku/heroku-buildpack-python/pull/1094#issuecomment-778201602

However thank you for suggesting this change, and please do open issues for any other ideas you may have :-)

timmygee:
To you decision to not support this, I respectfully object. Complex Django applications such as my own, rely on dependency resolution as provided by `Pipenv.lock`. It also requires the python [cryptography](https://github.com/pyca/cryptography) module, which itself requires a recent version of pip to actually build and install. pip version 9 does not cut it. The fact that this buildpack only provides pip version 9 for projects built out with pipenv, when the latest version of pip is version 20+, is not great.

I'd switch to poetry, which seems to be preferred over pipenv these days but that's not supported by this buildpack at all.

I can now literally only get my project deployed using @alan-eu's forked version of this buildpack.

I know I'm not the only one who has this problem. Anyone who uses pipenv and packages that require a modern version of pip will also have this issue. Please support the python community when it comes to this.

tolomea:
@timmygee The pull @edmorley linked updates both Pip and Pipenv to current versions.

timmygee:
[...]

Issue #987, opened 2020-06-25
Failed install with outdated pipenv version


nikolaik:
I think this build fails because of a bug fixed in a more recent pipenv version (latest is at 2020.6.2 at the moment)

See attached build log below:
```
-----> Found python-3.7.3, removing
cp: cannot stat '/tmp/build_d0defce75280ad585c4ef561790992d0/requirements.txt': No such file or directory
-----> Installing python-3.8.3
-----> Installing pip
-----> Installing dependencies with Pipenv 2018.5.18…
       Installing dependencies from Pipfile.lock (ddf461)…
       An error occurred while installing pycryptodome==3.9.*! Will try again.
       Installing initially–failed dependencies…
       Collecting pycryptodome==3.9.* 
         Using cached https://files.pythonhosted.org/packages/26/29/73158aec5b839234920c934ca6b81af260eaedd1f93ce341e05fa32c89b4/pycryptodome-3.9.8-cp38-cp38-manylinux1_x86_64.whl
       
       THESE PACKAGES DO NOT MATCH THE HASHES FROM Pipfile.lock!. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
           pycryptodome==3.9.* from https://files.pythonhosted.org/packages/26/29/73158aec5b839234920c934ca6b81af260eaedd1f93ce341e05fa32c89b4/pycryptodome-3.9.8-cp38-cp38-manylinux1_x86_64.whl#sha256=dd302b6ae3965afeb5ef1b0d92486f986c0e65183cd7835973f0b593800590e6 (from -r /tmp/pipenv-f08ruld6-requirements/pipenv-d2z_bfac-requirement.txt (line 1)):
               Expected sha256 07024fc364869eae8d6ac0d316e089956e6aeffe42dbdcf44fe1320d96becf7f
               Expected     or 09b6d6bcc01a4eb1a2b4deeff5aa602a108ec5aed8ac75ae554f97d1d7f0a5ad
               Expected     or 0e10f352ccbbcb5bb2dc4ecaf106564e65702a717d72ab260f9ac4c19753cfc2
               Expected     or 1f4752186298caf2e9ff5354f2e694d607ca7342aa313a62005235d46e28cf04
               Expected     or 2fbc472e0b567318fe2052281d5a8c0ae70099b446679815f655e9fbc18c3a65
[...]

PR #1094, opened 2020-10-08
manual override of pip version for pipenv users (PR)

vityas:
Allow for a manual override of pip version for pipenv users (PR)
https://github.com/heroku/heroku-buildpack-python/issues/1093

### DIFF
--- a/bin/steps/python
+++ b/bin/steps/python
@@ -179,7 +179,12 @@ if [[ -f "$BUILD_DIR/Pipfile" ]]; then
   # The buildpack is pinned to old pipenv, which requires older pip.
   # Pip 9.0.2 doesn't support installing itself from a wheel, so we have to use split
   # versions here (ie: installer pip version different from target pip version).
-  PIP_VERSION='9.0.2'
+  if [ -z "$PIP_VERSION_OVERRIDE" ]
+    then
+      PIP_VERSION='9.0.2'
+    else
+      PIP_VERSION=PIP_VERSION_OVERRIDE
+  fi
   PIP_TO_INSTALL="pip==${PIP_VERSION}"
 else
   PIP_TO_INSTALL="${PIP_WHEEL}"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 - Add support for Heroku-20 (#968).
-- allow pipenv users to specify pip version via `PIP_VERSION_OVERRIDE` variable (#1094)
+- Allow pipenv users to specify pip version via `PIP_VERSION_OVERRIDE` variable (#1094)
 
 ## v182 (2020-10-06)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Add support for Heroku-20 (#968).
+- allow pipenv users to specify pip version via `PIP_VERSION_OVERRIDE` variable (#1094)
 
 ## v182 (2020-10-06)
 


bin/steps/python
```diff
@@ -179,7 +179,12 @@ if [[ -f "$BUILD_DIR/Pipfile" ]]; then
   # The buildpack is pinned to old pipenv, which requires older pip.
   # Pip 9.0.2 doesn't support installing itself from a wheel, so we have to use split
   # versions here (ie: installer pip version different from target pip version).
-  PIP_VERSION='9.0.2'
+  if [ -z "$PIP_VERSION_OVERRIDE" ]
```
edmorley, line 182: Unfortunately this won't work, since `$PIP_VERSION_OVERRIDE` will never be set here, since app-provided env vars are passed in via `$ENV_DIR`.

edmorley (changes requested):
Hi! Thank you for the PR. I can empathise that being stuck on an old version of Pip if using Pipenv with this buildpack must be frustrating.

However I think it would be better to upgrade pipenv in this buildpack and then be able to remove this pipenv-pip-version special-casing entirely.

Allowing the pip version to be overridden opens the door for a whole lot of pain, since there is very tight integration between the buildpack and pip, meaning that even things like pip removing support for a particular CLI argument has the potential to cause breakage. (Or conversely, pip adds support for something new, the buildpack updates to that pip and starts using the new feature flag, only to have some apps break since the pip version they set N years ago doesn't support it.) This would result in an increase in support tickets, that are often hard to debug and time consuming.

I'd recommend:
1. Short term: Removing the `9.0.2` specification from a custom branch that you use on your apps (one of the reasons the buildpacks are open source is so that people can fork them if desired)
2. Medum/longer term: Exploring upgrading pipenv (I know this was ruled out in the past, but that was before the change of language owners, and I'm more open to the idea)

(Also, whatever approach is picked we always need tests with a change - in particular tests would have caught the fact that this PR doesn't currently work)

Anyway sorry not to have the answer you were hoping for - happy to chat more about this if you have any questions :-)

edmorley:
Hi! The pipenv version has just been updated in #1169. Some more context on that can be found here:
https://github.com/heroku/heroku-buildpack-python/issues/987#issuecomment-778151804

Thank you again for the PR, however supporting custom pipenv/pip versions is something that I believe will introduce more issues than it fixes for the reasons mentioned above (https://github.com/heroku/heroku-buildpack-python/pull/1094#pullrequestreview-506483377), so I won't be able to merge this.

To give a more concrete example of the types of problems we will encounter...

Imagine a user sets a custom version on their only (production) app, and a few months/years later start using a staging app, review apps or Heroku CI. They forgot they added the custom env var, so didn't set it on the staging app and/or apps.json config. This works fine initially, so long as the buildpack's default version (and CLI flags used) are compatible with their custom version. Later on the buildpack changes pip/pipenv version and/or CLI flags used, causing builds break in one of their environments but not the other. This is confusing for them to debug (and they may not be inclined to debug, particularly since nothing changed on their side) - so they open a support ticket saying "we haven't changed anything on our app for a while, the breakage must be something Heroku has done, please fix it". Now only 0.1% of customers may hit this case, however at cloud-scale this means days/weeks of numerous support tickets.

Another issue is test matrix permutations. As-is the test suite has to test multiple Heroku stacks, versions of Python, different package manages, etc. If we add support for say configuring pip and pipenv versions there are immediately 3 more permutations (both set to custom, only pipenv set to custom, only pip set to custom), even before we potentially test some a handful of different pipenv/pip versions that customers might be using.

The best approach here is for the buildpack to stay more up to date with pipenv versions, and then the need for customising version goes away.

Anyway hope that helps give some more context :-)

PR #1120, opened 2020-11-12
Remove env vars used by legacy build systems

edmorley:
These variables were used by Anvil and/or the legacy `slug_compiler` that predated cytokine.

I've not added a changelog entry since none were exported, so wouldn't have been available to subprocesses anyway.

Closes [W-8387606](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008VnOtIAK/view).

[skip changelog]

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -73,19 +73,13 @@ RECOMMENDED_PYTHON_VERSION=$DEFAULT_PYTHON_VERSION
 export PATH=$PATH:$ROOT_DIR/vendor/:$ROOT_DIR/vendor/pip-pop
 
 # Set environment variables if they weren't set by the platform.
-# Note: this is legacy, for a deprecated build system known as Anvil.
-# This can likely be removed, with caution.
-[ ! "$SLUG_ID" ] && SLUG_ID="defaultslug"
-[ ! "$REQUEST_ID" ] && REQUEST_ID=$SLUG_ID
 [ ! "$STACK" ] && STACK=$DEFAULT_PYTHON_STACK
 
 # Sanitize externally-provided environment variables:
 # The following environment variables are either problematic or simply unneccessary
 # for the buildpack to have knowledge of, so we unset them, to keep the environment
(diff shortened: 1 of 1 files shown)


bin/compile
```diff
@@ -73,19 +73,13 @@ RECOMMENDED_PYTHON_VERSION=$DEFAULT_PYTHON_VERSION
...
-# This can likely be removed, with caution.
-[ ! "$SLUG_ID" ] && SLUG_ID="defaultslug"
-[ ! "$REQUEST_ID" ] && REQUEST_ID=$SLUG_ID
 [ ! "$STACK" ] && STACK=$DEFAULT_PYTHON_STACK
 
 # Sanitize externally-provided environment variables:
 # The following environment variables are either problematic or simply unneccessary
 # for the buildpack to have knowledge of, so we unset them, to keep the environment
 # as clean and pristine as possible.
-unset GIT_DIR PYTHONHOME PYTHONPATH
```
dzuelke, line 86: `GIT_DIR` should be unset. If you `git push` to Heroku using SSH as the transport (as opposed to HTTPS), it'll lead to trouble later on if anything your buildpack does uses `git`.

See https://github.com/heroku/heroku-buildpack-php/commit/639f05141fc6138ae6ea04d3844d9de870dc5b4e

dzuelke (changes requested):

bin/compile
edmorley: Cytokine unsets it:
https://github.com/heroku/cytokine/blob/2357eb8e56e0f7d0bcc55e8b1d27af1cec3a73dc/buildpack/buildpack.go#L24-L32

(try it :-))

edmorley: Also the builds don't run on the git server (not since cedar original?). The source is packaged up, uploaded to the sources API, then as build triggered via the builds API.

dzuelke (approved):

Issue #1006, opened 2020-07-23
Editable installs don't work with setuptools 47.2.0+ and multi-buildpack


edmorley:
In https://heroku.support/890604 an app using pip editable installs combined with multiple buildpacks started seeing build failures when using newer setuptools.

The output from a reduced testcase looks like:

```
remote: Traceback (most recent call last):
remote:   File "/app/.heroku/python/bin/demo_say_hello", line 33, in <module>
remote:     sys.exit(load_entry_point('demobuildissue', 'console_scripts', 'demo_say_hello')())
remote:   File "/app/.heroku/python/bin/demo_say_hello", line 22, in importlib_load_entry_point
remote:     for entry_point in distribution(dist_name).entry_points
remote:   File "/app/.heroku/python/lib/python3.8/importlib/metadata.py", line 504, in distribution
remote:     return Distribution.from_name(distribution_name)
remote:   File "/app/.heroku/python/lib/python3.8/importlib/metadata.py", line 177, in from_name
remote:     raise PackageNotFoundError(name)
remote: importlib.metadata.PackageNotFoundError: demobuildissue
```

Investigating I found that:

- between setuptools v47.1.1 and 47.2.0 the wrapper scripts created for editable-install packages (only) were refactored to use a different module for resolving the target package (in https://github.com/pypa/setuptools/pull/2194)
- this new module no longer looks at the `.egg-link` files in `site-packages`, and only the standard Python import paths (incl `.pth` files)

The reason this causes breakage in the multi-buildpack case is:

- the build system runs builds in a different directory (`/tmp/build_*`) from which the app will be run at runtime (`/app`)
- much of Python + its tooling expects it's location not to change after install (ie: it's not relocatable) and as such the Python buildpack has to use several path rewriting hacks to make this work
- one such hack is rewriting the `.pth` files added to `site-packages` for editable installs (installs where the package is effectively symlinked from its source directory rather than being packaged and copied into site-packages) so that the absolute paths inside reference `/app` instead of the `/tmp/...` paths of `BUILD_DIR` 
- this rewrite occurs right at the end of the Python compile step, since at that point it's presumed all Python commands have been run (eg: the `bin/post_compile` hook has already been called)
[...]

PR #1130, opened 2020-12-08
Fix for Pipenv not caching dependencies

litchfield:
Having 'git' in one of your package names will prevent your deps from caching, resulting in painfully long build times. This is a simple fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fixed `pipenv` not caching dependencies when `git` mentioned in package names.
+
 ## v187 (2020-12-08)
 
 - Python 3.9.1 is now available (CPython) (#1127).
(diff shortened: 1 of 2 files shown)


edmorley (commented):
@litchfield Sorry for the delayed reply. I don't suppose you have a testcase `Pipfile` + `Pipfile.lock` I can use to try out the before/after?

I'm also curious about the choices behind the buildpack's existing behaviour to skip pipenv install. Specifically:
1. Is the speedup here from not having to install pipenv itself, or in being able to skip the `pipenv install` step?
2. Why does this code even need to check for the presence of `git` in `Pipfile.lock` at all? Surely the lockfile should have an explicit SHA256 and not a branch name, so be deterministic? If not, should pipenv change here?

litchfield:
Thanks. Reading the patch, this was a quick fix to tighten existing regex and get it working. But I agree, not sure why we would even check if we have a lock file. I don’t have a pip file handy but any with a git source reference should do it.

edmorley:
Reading through #656 and #654 it appears the issue is more to to with editable dependencies rather than git. I'm pretty sure this is due to the path rewriting hacks (#1006 has some background on these) meaning that editable installs aren't correctly restored for subsequent builds.

Hopefully in the future when we can remove the path rewriting the need to repeat the pipenv install may go away.

edmorley:
The fragile `git` check has been removed as of #1526. The `pipenv install` step is now run every time, however, this doesn't mean the cache isn't used. The `site-packages` directory is still cached - it will just now be up to Pipenv to determine whether the environment needs updating, or whether the install step can no-op.

PR #1147, opened 2021-01-10
Pass detection if a runtime.txt is present

jkutner:
This would be useful for apps that only need a Python runtime, and don't depend on any packages.

For example https://github.com/jkutner/web-tty-buildpack

### DIFF
--- a/bin/detect
+++ b/bin/detect
@@ -15,7 +15,7 @@
 BUILD_DIR=$1
 
 # Exit early if app is clearly not Python.
-if [ ! -f "$BUILD_DIR/requirements.txt" ] && [ ! -f "$BUILD_DIR/setup.py" ] && [ ! -f "$BUILD_DIR/Pipfile" ]; then
+if [ ! -f "$BUILD_DIR/requirements.txt" ] && [ ! -f "$BUILD_DIR/setup.py" ] && [ ! -f "$BUILD_DIR/Pipfile" ] && [ ! -f "$BUILD_DIR/runtime.txt" ]; then
   exit 1
 fi
 


edmorley:
@jkutner Hi! I've been thinking about doing this for a while, and have a fixture in the new Hatchet tests I'm writing to cover the runtime-only case, which I was going to land as failing detect initially, and revisit when the implementation changed.

This change will mean that any project that has a `runtime.txt` but whose primary language is one of those after Python in the default detect order () will now fail unless explicit buildpacks are specified. However I think that case is likely rare for `runtime.txt`, given it's mostly a heroku-only concept, so less likely to be in the repo root than say a `requirements.txt` (were we not already supporting the requirements.txt only case for example).

That said, with this change the compile still fails:

```
-----> Python app detected
cp: cannot stat '/build/requirements.txt': No such file or directory
-----> Installing python-3.9.1
-----> Installing pip 20.1.1, setuptools 47.1.1 and wheel 0.34.2
-----> Installing SQLite3
-----> Installing requirements with pip
       ERROR: File "setup.py" not found. Directory cannot be installed in editable mode: /build
```

I've filed W-8720280 so make sure this doesn't get forgotten.

edmorley (changes requested):
Compile fails with:
`ERROR: File "setup.py" not found. Directory cannot be installed in editable mode: /build`

edmorley:
Closing since tests are failing due to needing further changes.

This is something I'd like to support (tracked at GUS-W-8720280), though it's likely it will first feature in the future native Python CNB before the classic buildpack.

PR #1178, opened 2021-02-19
Add support for Python 3.8.8 and 3.9.2

edmorley:
https://blog.python.org/2021/02/python-392-and-388-are-now-available.html
https://www.python.org/downloads/release/python-388/
https://www.python.org/downloads/release/python-392/

Closes GUS-W-8159179.
Closes GUS-W-8513391.


edmorley:
CI is failing because the binaries on S3 don't exist yet (since the release isn't due until later today), however it would be great to get this reviewed prior to then, so that I can release these security updates as soon as they are available.

dzuelke (approved):

PR #1187, opened 2021-04-01
Use Python 3.9 as the default Python version for new apps

edmorley:
Previously if a new app did not specify a Python version via a `runtime.txt` or the appropriate `Pipfile` config option, then the buildpack would default to Python 3.6.13.

Now the buildpack will use Python 3.9.2.

This does not affect:
* existing apps that have already had a build using the previous default version (since that version is remembered from one build to the next, via the "sticky versions" feature).

Use-cases involving new apps that aren't compatible with Python 3.9 will now need to specify Python 3.6 explicitly:
https://devcenter.heroku.com/articles/python-support#specifying-a-python-version

Closes GUS-W-7686321.


schneems (approved):

edmorley:
https://devcenter.heroku.com/changelog-items/2090

PR #1296, opened 2022-03-22
Use Python 3.10 as the default Python version for new apps

edmorley:
Previously if a new app did not specify a Python version via a `runtime.txt` or the appropriate `Pipfile` config option, then the buildpack would default to the latest Python 3.9 release (currently Python 3.9.11).

Now the buildpack will use the latest Python 3.10 release (currently Python 3.10.3).

This does not affect:
* existing apps that have already had a build using the previous default version (since that version is remembered from one build to the next, via the "sticky versions" feature).

Use-cases involving new apps that aren't compatible with Python 3.10 will now need to specify Python 3.9 explicitly:
https://devcenter.heroku.com/articles/python-support#specifying-a-python-version

However most apps should be compatible, since Python 3.10 was released some time ago (2021-10-04), and has been happily used by the [python-getting-started](https://github.com/heroku/python-getting-started) project since 2021-10-29.

GUS-W-10881777.


dzuelke (approved):

edmorley:
https://devcenter.heroku.com/changelog-items/2371

Issue #1225, opened 2021-07-29
how to use --enable-shared ?


buncis:
I've tried to fork this repository and add `--enable-shared` to this line

https://github.com/heroku/heroku-buildpack-python/blob/main/builds/runtimes/python3#L26

```
./configure --prefix=$OUT_PREFIX --with-ensurepip=no --enable-shared
make
make install
```

but it not works my heroku builded without using the python that have enable-shared 
even I think heroku build the python without using my buildpacks at all 
any guidance how to use python3 that compiled using --enable-shared ?

edmorley:
@buncis Hi! The scripts under `builds/` in this repository are not run when each app is built using the buildpack, but separately when releasing new Python versions to S3. You will need to run the compile scripts uploading to your own S3 bucket ([build docs]()), then use `BUILDPACK_S3_BASE_URL` to point at that bucket (after mirroring the other contents eg pip wheels). `BUILDPACK_S3_BASE_URL` cannot be set as an app env var, and will need to be set via an earlier buildpack's export script (or just fork the buildpack).

buncis:
@edmorley thanks for the quick response

I see, so the correct way to enable-shared is just like this fork right ?

https://github.com/ReforgeHQ/heroku-buildpack-python/commit/813825b1d886aa78f70070cedcae3aeb1ea8fe46

I've tried that way too, I fork that fork and change the line to my desired python version but somehow its not working and I forced to use heroku-18 stack for that fork

ok please bear with me

so what I need to do is

[ ] compile the python myself with --enable-shared configuration
[ ] upload the compile result to amazon s3
[ ] fork this heroku-buildpack-python and change the `BUILDPACK_S3_BASE_URL` to my own s3

edmorley:
> so what I need to do is

Yeah that's one approach.

Another would be to just compile during the build, and store the result in the cache, so you don't need to mess about with S3.

Example at:
https://github.com/heroku/heroku-buildpack-python/issues/488#issuecomment-659579578

(Though note this is all unsupported and you'll unfortunately mostly need to figure this out yourself)

buncis:
I think I will follow that #488 approach since its almost have same use case and simpler

but I don't know where to put the `post_compile` scripts

did this is the correct file ?
https://github.com/heroku/heroku-buildpack-python/blob/f3abce6ac660a009c7ffac95d5b86d231555c28a/bin/steps/hooks/post_compile

yeah If I could pass this 
[...]

Issue #665, opened 2018-03-19
add --enable-shared option


strychnide:
this enables cython support in some (all?) cases, and takes only an additional --enable-shared option while building
probably a separated runtime option with the flag enabled (even only for some versions) would be helpful

kennethreitz:
We used to do this, but it caused more confusion (and was used so little) than anything else.

PR #1320, opened 2022-05-03
Binary build script: Use shared builds + LTO for Python 3.10

### LINKED ISSUE #243: Impossible to get runtime with shared library on heroku?
I need to get a version of Python working with shared library support. I've tried a number of things:

  * Use this `heroku-buildpack-python` and specified `python-2.7.8-shared` in `runtime.txt`. This gives me 2.7.8 allright. But in `/app/.heroku/python/lib` I still find `libpython2.7.a`, so it's useless.
  * Specify the other "official" shared release: `python-2.7.7-shared`. Gives me an error message `"Requested runtime (python-2.7.7-shared) is not available for this stack (cedar-14)."`
  * Forked this buildpack, add my own version `python-2.7.10-shared` with the `--enable-shared` compilation flag. Then I told Heroku to use my own forked buildpack. Again when I deploy, it gives me an error message `"Requested runtime (python-2.7.7-shared) is not available for this stack (cedar-14)."`

I'm also not sure whether Heroku actually downloads and compiles a new Python runtime, or takes a runtime from some pre compiled cache?

edmorley:
Shared builds are beneficial for a number of reasons:
- Reduces the size of the build, since it avoids the duplication between the Python binary and the static library.  For example, for Python 3.10.x on Heroku-20, the output size drops from 206MB -> 90MB.
- Permits use-cases that only work with the shared Python library, and not the static library (such as `pycall.rb` or `PyO3`).
- More consistent with the official Python Docker images and other distributions.

However, shared builds are slower unless `no-semantic-interposition`and LTO is used:
https://fedoraproject.org/wiki/Changes/PythonNoSemanticInterpositionSpeedup

It's only as of Python 3.10 that `no-semantic-interposition` is enabled by default, so we only use shared builds on Python 3.10+ to avoid needing to override the default compiler flags.

Build size reductions:

```
$ diff -U0 <(cd with-system-expat/ && du -Sh) <(cd shared-lto/ && du -Sh)
--- /dev/fd/63
+++ /dev/fd/62
@@ -1 +1 @@
-24M    ./bin
+56K    ./bin
@@ -78 +78 @@
-60M    ./lib/python3.10/config-3.10-x86_64-linux-gnu
+144K   ./lib/python3.10/config-3.10-x86_64-linux-gnu
@@ -81 +81 @@
-20M    ./lib/python3.10/lib-dynload
+18M    ./lib/python3.10/lib-dynload
@@ -103 +103 @@
-59M    ./lib
+26M    ./lib
```

```
$ diff -U0 <(cd with-system-expat/ && du -sh) <(cd shared-lto/ && du -sh)
--- /dev/fd/63
+++ /dev/fd/62
@@ -1 +1 @@
-206M   .
+90M    .
```

Also adds a `du` at the end of the build scripts, to make it easier to see output sizes.

Note: This change will only take effect for future Python version releases (or future Heroku stacks) - existing Python binaries are not being recompiled.

Configure docs:
https://docs.python.org/3/using/configure.html#cmdoption-enable-shared
https://docs.python.org/3/using/configure.html#cmdoption-with-lto
https://docs.python.org/3/using/configure.html#cmdoption-without-static-libpython

See also:
https://github.com/docker-library/python/blob/1cf43e70e45843c70909a5f914c3c6d0f85fc200/Dockerfile-linux.template#L154-L159
https://github.com/docker-library/python/issues/501
https://github.com/docker-library/python/pull/502
https://github.com/docker-library/python/pull/660
https://github.com/python/cpython/pull/24418

Closes #243, #665, #1225.
GUS-W-10989125.


doriskwan (approved):

edmorley:
For a summary of the combined Python runtime size reductions from this and related PRs, see:
https://github.com/heroku/heroku-buildpack-python/pull/1322#issuecomment-1120943848

PR #1326, opened 2022-05-17
Add support for Python 3.9.13

edmorley:
https://blog.python.org/2022/05/python-3913-is-now-available.html
https://www.python.org/downloads/release/python-3913/

This release is the first that makes use of the updated build scripts, including now using a shared build, stripped binaries and other size-reduction improvements. 

Note: The GPG verification added as part of that build script refactoring work is failing on Heroku-18 due to issues connecting to the keyserver (it had been tested on Heroku-22 and Heroku-20 only), and so has been skipped for now since there doesn't appear to be a quick fix (and likely not worth fixing longer term, given Heroku-18 will be deprecated shortly).

GUS-W-10333173.


jonnymacs (approved):
👍

PR #1350, opened 2022-08-15
Skip check for requirement.txt changes when using Pipenv

### LINKED ISSUE #1270: Should I have `requirements.txt` when using pipenv?
I have a python project and store the dependencies in Pipfile.lock. Obviously the buildpack understands the Pipfile.lock as it installs the dependencies and everything works okay. As a result I do not use my requirements.txt file. Once I removed it, the build continued to work as expected, however now there is an error thrown:

```
cp: cannot stat '/tmp/build_a123c0456/requirements.txt': No such file or directory
```

Is that the expected behavior and should I have a requirements.txt file? If so, what should be it's content?

edmorley:
Previously Pipenv builds would include a stray error message in the log output, that didn't affect the result of the build, but added noise to the logs and gave the false impression the build was broken in some way:

```
cp: cannot stat '/tmp/build_a123c0456/requirements.txt': No such file or directory
```

This originated from the check of `requirements.txt` changes compared to the copy of the file seen in the last build. This check is only necessary when using Pip (and not Pipenv), so should always have been skipped.

Now that the check is fixed, the disabling of bash exit on error mode for part of `bin/steps/python` can more easily be removed, increasing coverage of such issues in the future.

Fixes #1270.
GUS-W-7924941.


bin/steps/python
```diff
@@ -1,7 +1,11 @@
 #!/usr/bin/env bash
 
-set +e
-runtime-fixer runtime.txt
+# TODO: Replace runtime-fixer with a bash equivalent, since currently on rebuilds it
+# gets run with the cached Python install (rather than system Python), which means it
+# fails to run with glibc errors on stack downgrades (see spec/hatchet/stack_spec.rb).
+# To work around this for now, non-zero exit codes are ignored (most runtime.txt files
+# don't need fixing up, so the script not running won't normally matter).
+runtime-fixer runtime.txt || true
```
edmorley, line 8: Rather than use `set +e` around just this section (which is the last remaining misbehaving part of this file, now that the other error below was fixed in this PR), I've switched to the `|| true` approach for clarity.

doriskwan (approved):

PR #1344, opened 2022-08-14
Update pip to 22.2.2 and setuptools to 63.4.3

edmorley:
Updates (for Python 3.7+ only):
- pip from 22.1.2 to 22.2.2
- setuptools from 60.10.0 to 63.4.3

Changelogs:
https://pip.pypa.io/en/stable/news/#v22-2-2
https://setuptools.pypa.io/en/latest/history.html#v63-4-3

Full changes:
https://github.com/pypa/pip/compare/21.1.2...22.2.2
https://github.com/pypa/setuptools/compare/v60.10.0...v63.4.3

Pip wheel uploaded using:

```
$ pip download --no-cache pip==22.2.2
...
Saved ./pip-22.2.2-py3-none-any.whl
Successfully downloaded pip
$ aws s3 sync . s3://heroku-buildpack-python/common/ --exclude "*" --include "pip-*.whl" --profile python --dryrun
(dryrun) upload: ./pip-22.2.2-py3-none-any.whl to s3://heroku-buildpack-python/common/pip-22.2.2-py3-none-any.whl
$ aws s3 sync . s3://heroku-buildpack-python/common/ --exclude "*" --include "pip-*.whl" --profile python
upload: ./pip-22.2.2-py3-none-any.whl to s3://heroku-buildpack-python/common/pip-22.2.2-py3-none-any.whl
```

GUS-W-11592987.
GUS-W-11592988.

PR #1355, opened 2022-08-16
Support setuptools v64's PEP660 editable install mode

mrcljx:
We were running into `ModuleNotFoundError` errors when trying
to run our application. After some investigation we noticed
that newer setuptools places new files `site-packages` when
using editable installs (I guess only in some situations):

```
mylib-0.0.0.dist-info
__editable__.mylib-0.0.0.pth
__editable___mylib_0_0_0_finder.py
```

The `*.pth` file in that case contains this stub:

```
import __editable___mylib_0_0_0_finder; __editable___mylib_0_0_0_finder.install()
```

while the `py` module is generated from a [template][1] and contains this line:

```
MAPPING = {'mylib': '/tmp/build_892299be/py-packages/mylib/src/mylib'}
```

This all seems to be related to setuptools 63 and might be mitigated
by the [compat mode in 64][2]. I assume we just started to see this
issue because of the [upgrade from 60 to 63][3] yesterday.

This PR updates the buildpack to also patch the paths in those files.
Please let me know if you have any recommendations for tests to add.

[1]: https://github.com/pypa/setuptools/blob/0a1906819c568ceb3ffccc61d531fecd6b8662de/setuptools/command/editable_wheel.py#L740
[2]: https://github.com/pypa/setuptools/pull/3484
[3]: https://github.com/heroku/heroku-buildpack-python/commit/4646cffde336340c44b88adc499633df4f832c5e

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -323,7 +323,7 @@ EOT
 # - lib-python/3/site-packages/ (PyPy)
 if [[ "${BUILD_DIR}" != "/app" ]]; then
   cat <<EOT >> "$PROFILE_PATH"
-find .heroku/python/lib*/*/site-packages/ -type f -and \( -name '*.egg-link' -or -name '*.pth' \) -exec sed -i -e 's#${BUILD_DIR}#/app#' {} \+
+find .heroku/python/lib*/*/site-packages/ -type f -and \( -name '*.egg-link' -or -name '*.pth' -or -name '__editable___*_finder.py' \) -exec sed -i -e 's#${BUILD_DIR}#/app#' {} \+
 EOT
 fi
 
--- a/spec/fixtures/requirements_editable/bin/test-entrypoints
+++ b/spec/fixtures/requirements_editable/bin/test-entrypoints
@@ -2,8 +2,8 @@
 
 set -euo pipefail
 
-# List the filenames and contents of all .egg-link and .pth files in site-packages.
-find .heroku/python/lib*/*/site-packages/ -type f -and \( -name '*.egg-link' -or -name '*.pth' \) | sort | xargs -exec tail -n +1
+# List the filenames and contents of all .egg-link, .pth, and finder files in site-packages.
+find .heroku/python/lib*/*/site-packages/ -type f -and \( -name '*.egg-link' -or -name '*.pth' -or -name '__editable___*_finder.py' \) | sort | xargs -exec tail -n +1
 echo
 
 echo -n "Running entrypoint for the local package: "


edmorley:
@mrcljx Hi! Thank you for the write up and opening a PR :-) 

Would I be correct in thinking that the package being installed in editable mode has a `pyproject.toml` so is being installed via the new Pip build isolation mode? If so, I believe the cause of this is that in isolation mode, the latest version of setuptools is pulled in (via the build backend dependency ranges specified in `pyproject.toml`, unless they explicitly pin the setuptools version). 

If that's the case, this may even be unrelated to #1344 and instead just have been caused by setuptools v64 having been released around the same time (11th Aug):
https://setuptools.pypa.io/en/latest/history.html#v64-0-0

Either way, it seems we'll need this change to adapt for setuptools's new editable mode. I'll see if I can come up with a test.

mrcljx:
> If that's the case, this may even be unrelated to #1344 and instead just have been caused by setuptools v64 having been released around the same time (11th Aug): [setuptools.pypa.io/en/latest/history.html#v64-0-0](https://setuptools.pypa.io/en/latest/history.html#v64-0-0)

That would actually make a lot of sense as we started seeing failures last Thu/Fri (so around Aug 11) and not just yesterday - the patch here didn't seem to fully solve it yet for us either.

mrcljx:
Thanks for the hint! I looked a bit further into it: so these finder files are really only created when setuptools comes to the conclusion that the project setup is not trivial. After deleting `setup.py` and fully embracing `pyproject.toml` the `*.pth` files actually became simple paths again (and no `_finder.py` files were emitted).

edmorley:
> so these finder files are really only created when setuptools comes to the conclusion that the project setup is not trivial

Yeah - looks like for this issue to occur, all of the following has to be true:
- the package is being installed as an editable install
- the package has a `pyproject.toml` with a `build-system` table defined with the `setuptools` version range set so that setuptools v64+ gets pulled in (which includes the new PEP660 editable install feature rather than legacy editable mode)
- the install is being performed in "lenient" editable mode (the current setuptools default), not "strict" editable mode
- the logic [here](https://github.com/pypa/setuptools/blob/d03da04e024ad4289342077eef6de40013630a44/setuptools/command/editable_wheel.py#L359-L368) decides the project layout is complex enough, that the Python finder scripts are also created, rather than just using the static `.pth` file

It seems like workarounds (until this is merged/released), are:
1. Adjust the package config so setuptools no longer deems the config as "complex" (eg by removing setup.py entirely, as you did)
2. Use "strict" editable install mode instead (I'm presuming; I've not tested)
3. Switch back to legacy editable behaviour via the env var `SETUPTOOLS_ENABLE_FEATURE=legacy-editable`
4. Use `setuptools<64` in the `build-system` table in `pyproject.toml` to force the use of older setuptools

edmorley:
@mrcljx Thank you for the PR! Our CI needs secrets to run (so doesn't work from forks), and I also needed to add tests and changelog entry - so I had to merge this via a separate PR, but I credited you in the PR description and via the `Co-authored-by:` commit metadata (which GitHub uses to populate the attribution on the commit: b2dfe7397bf2f57b70799954277dd01a9c034432). I'll release this change shortly :-)

mrcljx:
Thanks @edmorley - appreciate you taking over.

PR #1363, opened 2022-09-21
Provision a Heroku Postgres DB in fewer cases

### LINKED ISSUE #1067: Skip the default heroku-postgresql addon if a Postgres DB driver isn't installed
Currently the Python buildpack's `bin/release` step sets a default `heroku-postgresql` addon if the app has a `manage.py` script.

However for use-cases where either no database is being used, or a non-Postgres DB is used, this default addon at best is unused and at worst causes confusion.

To resolve this the buildpack could check that the installed Python packages list includes `psycopg2`/...

This would be similar to the approach taken by the Ruby buildpack.

For more background see the discussion in:
https://github.com/heroku/heroku-buildpack-python/pull/1056

edmorley:
As part of the Buildpack API, a buildpack can request the provisioning of addons on the first build of an app via the `bin/release` payload. (This feature does not apply to subsequent builds of existing apps.)

Previously the Python buildpack requested a Heroku Postgres DB be provisioned if the app source contained a `manage.py` file in the source root or any subdirectory up to three levels deep.

This meant a Postgres DB was provisioned even in cases where the app wasn't a Django app (but just happened to have a `manage.py` script) or used an alternative DB backend (eg MySQL), or didn't use a DB at all.

Now, the DB is only provisioned if all of the following are true:
- a `manage.py` file exists in the root of the app source
- `django` is installed
- `psycopg2` is installed

This ensures it's still provisioned in the Python getting-started guide case (and any other similar cases), but not in the other scenarios mentioned above.

Fixes #1067.
GUS-W-8098791.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- The Heroku Postgres database auto-provisioning feature now provisions a DB in fewer cases ([#1363](https://github.com/heroku/heroku-buildpack-python/pull/1363)).
 
 ## v218 (2022-09-07)
 
--- a/bin/release
+++ b/bin/release
@@ -1,21 +1,17 @@
 #!/usr/bin/env bash
 # bin/release <build-dir>
 
+set -euo pipefail
+
 BUILD_DIR=$1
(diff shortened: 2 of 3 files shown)


Malax (approved):

edmorley:
https://devcenter.heroku.com/changelog-items/2471

PR #1364, opened 2022-09-22
Drop support for Python 2.7, 3.4, 3.5 and PyPy

edmorley:
Python 2.7, 3.4 and 3.5:
- are end-of-life upstream, so no longer receiving security updates making them insecure to use
- have previously been deprecated on the platform, with deprecation warnings
- have extremely low usage (Python 3.4 and 3.5 combined account for < 0.01% of builds, and Python 2.7 is < 0.3%)
- are only available on Heroku-18, which itself is deprecated

For PyPy:
- it was only ever a beta feature, and has been undocumented for some time
- the versions available on Heroku are outdated and insecure
- it has extremely low usage (<0.01% of builds)
- it's already only available on older stacks (Heroku-18 and Heroku-20)
- in general, only advanced users should be using PyPy as it comes with a number of caveats, and depending on the use-case may not offer enough of a performance improvement (particularly over newer CPython, which has been getting faster) to be worth the trade-offs.

Users wanting to continue using the above versions can:
- Pin to an older buildpack release, using the buildpack URL of `https://github.com/heroku/heroku-buildpack-python#v218`
- (For PyPy) Switch to the container stack and use the upstream [official PyPy Docker image](https://hub.docker.com/_/pypy/) instead.

Otherwise the suggested migration paths are:
- (For older CPython) Upgrade to a newer Python version by updating `runtime.txt` or `Pipfile` + `Pipfile.lock`.
- (For PyPy) Switch to CPython by updating `runtime.txt` to specify `python-X.Y.Z` instead of `pypy*`.

GUS-W-11796962.
GUS-W-11796984.
GUS-W-11796975.


colincasey (approved):

edmorley:
Added docs to the Python 2.7 EOL FAQ about how to pin to the previous buildpack release, to continue building with Python 2:
https://devcenter.heroku.com/articles/python-2-7-eol-faq#what-heroku-buildpack-versions-support-python-2

edmorley:
https://devcenter.heroku.com/changelog-items/2472
https://devcenter.heroku.com/changelog-items/2473

PR #1375, opened 2022-10-04
Fix automatic provisioning of Postgres DB addons

### LINKED ISSUE #1374: Simple Django project deploys successfully on v218, fails on v219 and following
Hi, I have a [simple Django project](https://github.com/ehmatthes/learning_log_2e_django40) that I've used to teach deployment on Heroku for years. Deployments have stopped working recently, and I believe it's related to the [new approach](https://github.com/heroku/heroku-buildpack-python/pull/1363) to deciding whether to auto-provision a Heroku Postgres database.

Here's an [issue](https://github.com/ehmatthes/learning_log_2e_django40/issues/1) detailing everything I've tried in an attempt to troubleshoot deployment. I am not an expert in Heroku buildpacks, but I did make a [demo buildpack](https://github.com/ehmatthes/heroku-buildpack-python/tree/my_autoconfigure_101721) a couple years ago that automated Django deployments for simple projects, so I have some idea how they work.

This project meets all three listed requirements for having a Heroku Postgres database auto-provisioned: it has a manage.py file in the root directory, and both `django` and `psycopg2` are listed in requirements.txt. I forked the current version of this buildpack, and added some [diagnostics](https://github.com/ehmatthes/heroku-buildpack-python/blob/main/bin/compile#L275) in `bin/compile` after `collectstatic` is run, which shows that the project does pass the three tests as written in `bin/release`.

The project deploys with the new buildpack, but it does not have a Postgres database provisioned. All pages that don't require a database work; pages that require a database fail. I have deployed to free dynos and to paid dynos, without seeing any change in behavior. 

- Is there any obvious reason this project fails to have a Postgres database auto-provisioned despite seeming to meet the three requirements?
- Where is the `bin/release` script called from? I tried to move my diagnostics closer to the release phase, but couldn't find where the release script is called, so placing it after `collectstatic` was the best I could do.

Thanks.

edmorley:
In #1363 the logic for determining when to automatically provision a Heroku Postgres database addon was adjusted to reduce false positives.

Unfortunately, this broke provisioning in all cases, since it turns out that the build system does not source the `export` script prior to running `bin/release` (like it does when running subsequent buildpacks), and so the buildpack configured env vars (such as `PATH`) are not set.

This meant that when `bin/release` used the `is_module_available` utility, it was calling system Python (where the Django and psycopg2 packages were not installed) and so always failing the conditional.

Unfortunately we do not currently have an easy way to test `bin/release`, otherwise a test case would have been added in the original PR.

I had manually tested the original PR using the Python getting started guide, but had only checked via accessing the app and trying to run the Django `./manage.py migrate` command. However it seems the Django config for that project silently falls back to sqlite if `DATABASE_URL` is not set - so the project (and the migrate command) still worked.

For this PR I have manually tested to ensure the addon itself appears on the app.

Fix #1374.
GUS-W-11851647.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Fix automatic provisioning of Postgres DB addons ([#1375](https://github.com/heroku/heroku-buildpack-python/pull/1375)).
 
 ## v220 (2022-09-28)
 
--- a/bin/release
+++ b/bin/release
@@ -4,10 +4,20 @@
 set -euo pipefail
 
 BUILD_DIR=$1
-BIN_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
+BUILDPACK_DIR=$(cd "$(dirname "$(dirname "${BASH_SOURCE[0]}")")" && pwd)
+
+# Unfortunately the build system doesn't source the `export` script before
+# running `bin/release`, so we have to do so manually to ensure the buildpack
+# Python is used by `is_module_available` instead of system Python.
+# We also have to disable Bash error on undefined variables, since not all env
+# vars used in the export script will be set by default (eg `LIBRARY_PATH`).
+set +u
+# shellcheck source=/dev/null
+source "${BUILDPACK_DIR}/export"
+set -u
 
 # shellcheck source=bin/utils
-source "$BIN_DIR/utils"
+source "${BUILDPACK_DIR}/bin/utils"
 
 if [[ -f "${BUILD_DIR}/manage.py" ]] && is_module_available 'django' && is_module_available 'psycopg2'; then
 cat <<EOF


edmorley:
I've also updated the Buildpack API docs to mention this caveat for `bin/release`:
https://devcenter.heroku.com/articles/buildpack-api#bin-release

doriskwan (approved):

ehmatthes:
Thanks for the prompt fix, and the thorough writeup. I hope you don't mind a  brief followup.

First, your description of the core issue made me feel better about not being able to find a fix myself. I had hoped to make a quick fix and file a PR instead of an issue, but it was deeper than I could dig in a reasonable time. Can you tell me where `bin/release` is called from? I searched this repo (code, not issues) for `release`, and couldn't find anywhere that it's called. Is it called from outside the buildpack?

Second, this surprised me: "Unfortunately we do not currently have an easy way to test bin/release, otherwise a test case would have been added in the original PR." I talk to a lot of Django teachers/mentors, and one of our shared frustrations about most platforms is that their demo Django project doesn't even use a database. But the python-getting-started repo does have a tiny model and a page that saves an instance of the model. Can't you write a test script that deploys this project, and tests that the [db](https://github.com/heroku/python-getting-started/blob/main/hello/views.py#L12) page is rendered successfully? It's an interesting test to write, when you're testing the full deployment process.

Heroku's gotten a bit of heat recently, but I'm really happy to see changes made that need to happen for long-term sustainability. Your writeup of issues like this are a great example of what corporate open source can look like.

Thanks again!

edmorley:
Hi! No problem :-)

> Can you tell me where `bin/release` is called from?

The Heroku build system itself calls `bin/release` (like it also calls `bin/detect` and `bin/compile`). Unfortunately the codebase for that isn't open source. 

Longer term, we're going to migrate from the current buildpacks/build system to Cloud Native Buildpacks (https://buildpacks.io), which can more easily be run locally (using the Pack CLI), and moving much of the build system logic out of Heroku-only codebases, which will make reproducing and testing things much easier.

> Can't you write a test script that deploys this project, and tests that the [db](https://github.com/heroku/python-getting-started/blob/main/hello/views.py#L12) page is rendered successfully?

Yes we could. The issue is that the `/db` route will still render fine with sqlite, which is what the getting started guide will fall back to if `DATABASE_URL` is not set, so the app will still appear functional. Arguably the getting started guide project should be changed so it doesn't do that, however that work is low priority and so was out of scope of the original PR here (which was requested urgently for the upcoming changes to Heroku plans). Also, there have been discussions about potentially removing the auto-addon provisioning feature entirely (in part since it existed from a pre-`app.json` era, plus also isn't supported by CNBs at least at the moment), so I was reluctant to spend too much time on a custom solution here.

Ideally the build system would report in the build log what the `bin/release` step requested, and then our buildpack integration tests could simply confirm via the build log output, which would have allowed for a much quicker/simpler test.

ehmatthes:
Thank you so much, this is really helpful. :)

PR #1379, opened 2022-10-24
Add support for Python 3.11

edmorley:
https://blog.python.org/2022/10/python-3110-is-now-available.html
https://www.python.org/downloads/release/python-3110/

The default Python version remains unchanged (at 3.10.x) for now.

Note: Pipenv emits a `RuntimeWarning` on Python 3.11 which is coming from its vendored copy of `attr`. This was fixed in `attr` v22.1:
https://github.com/python-attrs/attrs/commit/a18b3957d55c669de56799a9a72ea31febe93ce4

However pipenv currently uses an older `attr`:
https://github.com/pypa/pipenv/blob/v2022.10.12/pipenv/vendor/vendor.txt#L2

Once there is a new pipenv release that includes an updated `attr`, we can update to it to prevent the warning in the logs.

GUS-W-10333186.


Malax (approved):

edmorley:
```
$ make publish
Deploy as version: v222 [y/n]? y

Tagging commit 7dc6bdec681d0a57fa8399ca99f48b127e65053e with v222... 
Total 0 (delta 0), reused 0 (delta 0), pack-reused 0
To https://github.com/heroku/heroku-buildpack-python
 * [new tag]         v222 -> v222

Publishing to the buildpack registry...
Two-factor code: ********************************************
Publishing buildpack heroku/python... done
Publishing buildpack was successful

Version  Released At               Status
───────  ────────────────────────  ─────────
222      2022-10-25T08:25:42.979Z  published
```

https://devcenter.heroku.com/changelog-items/2489

PR #1396, opened 2023-01-06
Remove support for skipping check-changelog using the PR description

edmorley:
Since:
- We only ever use the `skip changelog` label method of skipping the check.
- Supporting skip-comments in the PR description means the check has to listen to the `edited` event type, which causes action-churn every time a PR description is edited.
- Having a clear single way to skip the check seems preferable to a grab-bag of approaches.


Malax (approved):
👍🏻

PR #1403, opened 2023-02-03
Update pipenv version to 2022.11.30

alex20465:
Not much to describe, this is just a singleliner with the latest stable pipenv version.

### DIFF
--- a/bin/steps/pipenv
+++ b/bin/steps/pipenv
@@ -42,7 +42,7 @@ if [ ! "$SKIP_PIPENV_INSTALL" ]; then
             mcount "buildvar.PIP_EXTRA_INDEX_URL"
         fi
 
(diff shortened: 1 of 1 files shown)


edmorley:
Hi! Thank you for the PR. 

Updating to newer pipenv is on my list for next week - several releases since the old `2020.11.15` version we're on have had regressions, which is why it hasn't been updated sooner, but things are looking better with newer pipenv versions, so hopefully safer to try now.

The CI for this repo needs credentials to run, so I'll have to update this from a non-forked PR. (I'll also update the tests that need updating for this change at the same time.)

edmorley (commented):
Thank you again for the PR! 

Closing this in favour of #1407 (since there is a newer pipenv available now, a few more changes were needed for tests etc + CI can't run from forks).

PR #1415, opened 2023-02-20
Drop support for Python 3.6

edmorley:
Since:
- It has been end of life upstream since 2021-12-23, so is no longer receiving security updates.
- The buildpack has been warning about its end of life in the build log output since 2022-08-17.
- The upgrade from Python 3.6 to Python 3.7+ in many cases requires no changes to an application at all, or else some small dependency updates (compared to say the upgrade from Python 2 to 3).
- It is already not available on Heroku-22 (due to it being incompatible with OpenSSL 3), so apps will need to switch eventually.
- Any apps that must stay on Python 3.6 short term, can pin to the previous buildpack version as a stop-gap.
   
GUS-W-12345793.


edmorley:
https://devcenter.heroku.com/changelog-items/2546

PR #1432, opened 2023-04-05
Migrate Python runtime builds to GitHub Actions

edmorley:
* Builds are now performed on GitHub Actions (which supports IP allowlisting), and can be triggered via the dispatch job workflow.
* The build process no longer uses the legacy [bob-builder](https://github.com/heroku-python/bob-builder) tool.
* The identical and thus redundant per-version scripts have been removed.
* The per-stack Dockerfiles have been consolidated.

GUS-W-12964868.


edmorley:
Dry run (ie build only, skipping the upload) runs for all major Python versions can be see here:
https://github.com/heroku/heroku-buildpack-python/actions/workflows/build_python_runtime.yml

builds/build_python_runtime.sh
```diff
@@ -2,47 +2,47 @@
...
+PYTHON_VERSION="${1:?"Error: The Python version to build must be specified as the first argument."}"
+PYTHON_MAJOR_VERSION="${PYTHON_VERSION%.*}"
+
+INSTALL_DIR="/app/.heroku/python"
+SRC_DIR="/tmp/src"
+ARCHIVES_DIR="/tmp/upload/${STACK}/runtimes"
+
+echo "Building Python ${PYTHON_VERSION} for ${STACK}..."
+
+if [[ "${STACK}" != "heroku-18" && "${STACK}" != "heroku-20" && "${PYTHON_MAJOR_VERSION}" == 3.[7-8] ]]; then
```
Malax, line 14: When I saw the conditional on the GHA workflow I was wondering if that'd be the best place to put it. Makes more sense now when it's checked here too. :)

builds/build_python_runtime.sh
```diff
@@ -2,47 +2,47 @@
...
+  echo "Error: Python ${PYTHON_MAJOR_VERSION} is only supported on Heroku-20 and older!" >&2
   exit 1
 fi
 
-# See: https://www.python.org/downloads/ -> "OpenPGP Public Keys"
-case "${VERSION}" in
-  3.1[0-1].*)
+# The release keys can be found on https://www.python.org/downloads/ -> "OpenPGP Public Keys".
+case "${PYTHON_MAJOR_VERSION}" in
```
Malax, line 20: I assume that this naming is correct and `3.10` et.al are considered "major versions" even if the second part is usually denoting a minor version.

Malax (approved):

PR #1437, opened 2023-04-11
Update setuptools to 67.6.1 and wheel to 0.40.0

edmorley:
This PR:
- Updates setuptools from 67.5.0 to 67.6.1.
- Updates wheel from 0.38.4 to 0.40.0.

Of note, the setuptools introduces more strict parsing of version specifiers (some invalid but previously accepted package version strings will now be rejected by setuptools, if they don't conform to PEP 440).

If errors are encountered installing packages due to invalid package version strings, then users will need to update to fixed versions of those packages, or else pin to a previous buildpack release until the affected package is fixed.

For more information, see:
- https://setuptools.pypa.io/en/stable/history.html#v66-0-0
- https://peps.python.org/pep-0440/

In addition, changes have been made to the way setuptools handles editable installs, that may affect some projects using editable mode (`-e`) depending on the project layout. See:
- https://setuptools.pypa.io/en/stable/history.html#v64-0-0
- https://setuptools.pypa.io/en/stable/userguide/development_mode.html

Full changelogs:
https://setuptools.pypa.io/en/stable/history.html#v67-6-1 https://wheel.readthedocs.io/en/stable/news.html

GUS-W-12346556.
GUS-W-13011106.


edmorley:
https://devcenter.heroku.com/changelog-items/2570

PR #1441, opened 2023-04-27
Update pip to 23.1.2 and setuptools to 67.7.2

edmorley:
- Updates pip from 23.0.1 to 23.1.2
- Updates setuptools from 67.6.1 to 67.7.2

See:
https://pip.pypa.io/en/stable/news/#v23-1-2
https://setuptools.pypa.io/en/stable/history.html#v67-7-2

The new pip version now outputs the origin of each requirement (eg the requirements file path which listed it). As such, the `pip install` invocations have also been adjusted to use a relative path to the app's requirements file, so that the log output doesn't end up including the full tmp directory path (eg `(/tmp/build_d73c77c7/requirements.txt`), which would make the logs more cluttered and harder to diff between builds.

Lastly, since #1442 has now landed, we no longer need to upload the new pip wheel version to S3, and so this pip upgrade just works as is.

GUS-W-12345615.
GUS-W-13108504.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Updated pip from 23.0.1 to 23.1.2. ([#1441](https://github.com/heroku/heroku-buildpack-python/pull/1441))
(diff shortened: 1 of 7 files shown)


Malax (approved):

edmorley:
https://devcenter.heroku.com/changelog-items/2579

PR #1508, opened 2023-11-08
Drop support for Python 3.7

edmorley:
Since:
- It has been end of life upstream [since 2023-06-27](https://devguide.python.org/versions/#supported-versions), so is no longer receiving security updates.
- The buildpack has been warning about its deprecation/end of life in the build log output since 2023-02-08 ([#1404](https://github.com/heroku/heroku-buildpack-python/pull/1404), [#1460](https://github.com/heroku/heroku-buildpack-python/pull/1460)), and its deprecation was previously announced on changelog: https://devcenter.heroku.com/changelog-items/2530
- The upgrade from Python 3.7 to Python 3.8+ in most cases requires no changes to an application at all (see [What’s New In Python 3.8](https://docs.python.org/release/3.8.18/whatsnew/3.8.html)), or else some small dependency updates (compared to say the upgrade from Python 2 to 3).
- Python 3.7 is not compatible with OpenSSL 3, and so only works on the Heroku-20 stack - which itself will be deprecated in 2024.
- Any apps that must stay on Python 3.7 short term, can [pin to the previous buildpack version](https://devcenter.heroku.com/articles/buildpacks#buildpack-references) (`v238`) as a stop-gap.

This also unblocks updating `setuptools` and `pipenv` in the buildpack, both of which have dropped support for Python 3.7 in recent versions.

Closes heroku/roadmap#198.
GUS-W-12345803.


Malax (approved):

edmorley:
https://devcenter.heroku.com/changelog-items/2719

PR #1547, opened 2024-03-05
Improve `WEB_CONCURRENCY` support

edmorley:
The Python buildpack, like several of the other languages buildpacks has for some time automatically set the `WEB_CONCURRENCY` environment variable at dyno boot (if it's not already set), based on the size of the Heroku dyno. The env var is then used by some Python web servers (such as Gunicorn and Uvicorn) to control the default number of server processes that they launch.

When the original Python buildpack implementation for this was written many years ago, there was not a way to determine dyno available memory vs the host memory (since `/sys/fs/cgroup/memory/memory.limit_in_bytes` did not exist). As such, the existing implementation relied upon a hardcoded mapping of known process limit values to dyno sizes:
https://devcenter.heroku.com/articles/limits#processes-threads

This mapping was fragile, since if the process limits ever changed in the future, or if new dyno types were added, then `WEB_CONCURRENCY` would not get set, or be set to an incorrect value.

In addition, the existing choice of concurrency values for Performance dynos (and their Private Space equivalents) was suboptimal, since on Performance-L dynos concurrency defaulted to 11, which is only 1.4 times the Performance-M's default concurrency of 8, even though the former has 4 times the number of CPU cores and 5.6 times the RAM.

As such, the buildpack now instead dynamically calculates the value for `WEB_CONCURRENCY` based on the dyno's actual specifications, by setting it to the lowest of either `<dyno available RAM in MB> / 256` or `<number of dyno CPU cores> * 2 + 1`. The former ensures each web server process has at least 256 MB RAM available to reduce the chance of OOM, and the latter is based upon benchmarking and the Gunicorn worker guidance here:
https://docs.gunicorn.org/en/latest/design.html#how-many-workers

This new implementation results in the following default concurrency values for each Heroku dyno size:

- `eco` / `basic` / `standard-1x`: 2 (unchanged)
- `standard-2x` / `private-s` / `shield-s`: 4 (unchanged)
- `performance-m` / `private-m` / `shield-m`: 5 (previously 8)
- `performance-l` / `private-l` / `shield-l`: 17 (previously 11)

To increase awareness of the change in defaults, and to make the buildpack's existing automatic configuration of `WEB_CONCURRENCY` less of a black box, the `.profile.d/` script now also prints memory/CPU/concurrency information to the app's logs (for web dynos only, to avoid breaking scripting use-cases). 

For example:

```
app[web.1]: Python buildpack: Detected 14336 MB available memory and 8 CPU cores.
app[web.1]: Python buildpack: Defaulting WEB_CONCURRENCY to 17 based on the number of CPU cores.
```

If your app is relying on the buildpack-set `WEB_CONCURRENCY` value, and you do not wish to use the new default concurrency values, then you can switch back to the previous values (or whatever value performs best in benchmarks of your app), by either:

- Setting `WEB_CONCURRENCY` explicitly as a config var on the app.
- Or, passing `--workers <num>` to Gunicorn/Uvicorn in your `Procfile` (which takes priority over any `WEB_CONCURRENCY` env var).

Lastly, integration tests have been added for the buildpack's `.profile.d/` scripts, since there were none before.

See:
https://devcenter.heroku.com/articles/config-vars
https://devcenter.heroku.com/articles/python-gunicorn
https://docs.gunicorn.org/en/latest/settings.html#workers
https://www.uvicorn.org/#command-line-options

GUS-W-14623334.
GUS-W-15109094.
GUS-W-15109115.
GUS-W-15131932.

### DIFF
--- a/spec/hatchet/profile_d_scripts_spec.rb
+++ b/spec/hatchet/profile_d_scripts_spec.rb
@@ -71,7 +71,7 @@ RSpec.describe '.profile.d/ scripts' do
 
       # Check WEB_CONCURRENCY support when using a Standard-1X dyno.
       # We set the process type to `web` so that we can test the web-dyno-only log output.
-      app.run_multi(list_concurrency_envs_cmd, heroku: { size: 'Standard-1X', type: 'web' }) do |output, _|
+      app.run_multi(list_concurrency_envs_cmd, heroku: { size: 'standard-1x', type: 'web' }) do |output, _|
         expect(output).to eq(<<~OUTPUT)
           Python buildpack: Detected 512 MB available memory and 8 CPU cores.
           Python buildpack: Defaulting WEB_CONCURRENCY to 2 based on the available memory.
@@ -81,7 +81,7 @@ RSpec.describe '.profile.d/ scripts' do
       end
 
       # Standard-2X
-      app.run_multi(list_concurrency_envs_cmd, heroku: { size: 'Standard-2X', type: 'web' }) do |output, _|
+      app.run_multi(list_concurrency_envs_cmd, heroku: { size: 'standard-2x', type: 'web' }) do |output, _|
         expect(output).to eq(<<~OUTPUT)
           Python buildpack: Detected 1024 MB available memory and 8 CPU cores.
           Python buildpack: Defaulting WEB_CONCURRENCY to 4 based on the available memory.
@@ -91,7 +91,7 @@ RSpec.describe '.profile.d/ scripts' do
       end
 
       # Performance-M
-      app.run_multi(list_concurrency_envs_cmd, heroku: { size: 'Performance-M', type: 'web' }) do |output, _|
+      app.run_multi(list_concurrency_envs_cmd, heroku: { size: 'performance-m', type: 'web' }) do |output, _|
         expect(output).to eq(<<~OUTPUT)
           Python buildpack: Detected 2560 MB available memory and 2 CPU cores.
           Python buildpack: Defaulting WEB_CONCURRENCY to 5 based on the number of CPU cores.
@@ -101,7 +101,7 @@ RSpec.describe '.profile.d/ scripts' do
       end
 
       # Performance-L
-      app.run_multi(list_concurrency_envs_cmd, heroku: { size: 'Performance-L', type: 'web' }) do |output, _|
+      app.run_multi(list_concurrency_envs_cmd, heroku: { size: 'performance-l', type: 'web' }) do |output, _|
         expect(output).to eq(<<~OUTPUT)
           Python buildpack: Detected 14336 MB available memory and 8 CPU cores.
           Python buildpack: Defaulting WEB_CONCURRENCY to 17 based on the number of CPU cores.
--- a/spec/hatchet/profile_d_scripts_spec.rb
+++ b/spec/hatchet/profile_d_scripts_spec.rb
@@ -1,0 +1,124 @@
+# frozen_string_literal: true
+
+require_relative '../spec_helper'
+
+RSpec.describe '.profile.d/ scripts' do
+  it 'sets the required run-time env vars' do
+    Hatchet::Runner.new('spec/fixtures/procfile', run_multi: true).deploy do |app|
+      # These are written as a single test to reduce end to end test time. This repo uses parallel_split_test,
+      # so we can't perform app setup in a `before(:all)` and have multiple tests run against the single app.
+
+      list_envs_cmd = 'env | sort | grep -vE "^(_|DYNO|PORT|PS1|SHLVL|TERM)="'
+
+      # Check all env vars are set correctly when there are no user-provided env vars.
+      # Also checks that the WEB_CONCURRENCY related log output is not shown for one-off dynos.
+      app.run_multi(list_envs_cmd) do |output, _|
+        expect(output).to eq(<<~OUTPUT)
+          DYNO_RAM=512
+          FORWARDED_ALLOW_IPS=*
+          GUNICORN_CMD_ARGS=--access-logfile -
+          HOME=/app
+          LANG=en_US.UTF-8
+          LD_LIBRARY_PATH=/app/.heroku/vendor/lib:/app/.heroku/python/lib:
+          LIBRARY_PATH=/app/.heroku/vendor/lib:/app/.heroku/python/lib:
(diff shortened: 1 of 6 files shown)


edmorley:
Note: This change will be a no-op for most non-Heroku usages of the buildpack (such as Dokku), since in those environments `WEB_CONCURRENCY` was not being set before (due to lack of the custom process limits), and still won't be now (since the buildpack skips configuring concurrency unless the `/sys/fs/cgroup/memory/memory.limit_in_bytes` file exists and contains a non-bogus value, which is only the case on Heroku, or when using cgroups v1 via older Docker with an explicit `--memory` limit set).

In the future we may choose to add support for cgroups v2's memory file too, however, enabling automatic `WEB_CONCURRENCY` in those non-Heroku environments would be a bigger breaking change, so has been deferred for now.

vendor/WEB_CONCURRENCY.sh
```diff
@@ -1,29 +1,78 @@
...
+
+# Prevents WEB_CONCURRENCY being set to zero if the environment is extremely memory constrained.
+if ((available_memory_in_mb < minimum_memory_per_process_in_mb)); then
+  max_concurrency_for_available_memory=1
+else
+  max_concurrency_for_available_memory=$((available_memory_in_mb / minimum_memory_per_process_in_mb))
+fi
+
+# This buildpack only supports being run on Heroku's base images, which all have nproc installed.
+cpu_cores=$(nproc)
```
dzuelke, line 61: Should this handle cases where `nproc` isn't there? Maybe fall back to `getconf -a | grep -E '_?NPROCESSORS_ONLN' | head -n1 | tr -s ' ' | cut -d" " -f2`?

spec/hatchet/profile_d_scripts_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+
+require_relative '../spec_helper'
+
+RSpec.describe '.profile.d/ scripts' do
+  it 'sets the required run-time env vars' do
+    Hatchet::Runner.new('spec/fixtures/procfile', run_multi: true).deploy do |app|
+      # These are written as a single test to reduce end to end test time. This repo uses parallel_split_test,
+      # so we can't perform app setup in a `before(:all)` and have multiple tests run against the single app.
+
+      list_envs_cmd = 'env | sort | grep -vE "^(_|DYNO|PORT|PS1|SHLVL|TERM)="'
```
dzuelke, line 11: I would whitelist the env vars, just in case it ever fails and for whatever reason exposes future variables that you don't want to be in CI run logs.

dzuelke (changes requested):

vendor/WEB_CONCURRENCY.sh
dzuelke: Alternatively, you could test this in a `heroku run` by prefixing your command with `nproc() { echo ''; }; export -f getconf;` if you wanted to.

edmorley: As the comment says (the comment was added because I knew you'd ask 😉), this buildpack only supports being run on our base images, and `nproc` is installed in all of them, so if someone tries running it in another image then they are well out of the supported use-case. 

Whilst we could technically have a `command -v nproc` conditional here, I decided against it since I felt we need to draw the line somewhere about what we can expect of the environment, given the buildpack is only intended for use on our images (and consumers like Dokku do use our images). As additional datapoints, I see the PHP buildpack doesn't check to see whether `python` is available [here](https://github.com/heroku/heroku-buildpack-php/blob/752a82c057408338293f25f4c423d53928390507/support/build/apache#L77) and several other buildpacks don't check whether `jq` is available etc (which similarly seem like reasonable assumptions, given that both `python` and `jq` are available in our base images).

That said, for completeness I had tested locally what the experience was when `nproc` was missing, and it results in `cpu_cores` being the empty string, which results in `WEB_CONCURRENCY=1` since the empty string is treated as zero in `$((cpu_cores * 2 + 1))`. So it at least falls back to something safe, and there will be clues for people to follow due to the `nproc: command not found` printed to stderr.

It's also worth noting that most non-Heroku environments won't even get as far as the `nproc` call, since they won't have the `/sys/fs/cgroup/memory/memory.limit_in_bytes` file, and so will have bailed earlier with the message:

```
Python buildpack: Skipping automatic configuration of WEB_CONCURRENCY since unable to determine available memory.
```

spec/hatchet/profile_d_scripts_spec.rb
edmorley: I agree there are trade-offs of both approaches. 

The reason I had chosen an exclusion list, was that I want to make sure any env vars added to the buildpack implementation in the future are caught by the test - and in the case of intentionally added env vars this will ensure test coverage is added, or in the case of accidentally exported internal vars, this will prevent shipping them by mistake.

The downsides to an exclusion list approach are the chance of (a) test failures/churn if the platform starts setting additional env vars, or (b) the chance of any new secret env vars. However, for (a) in general the approach I've taken for the test fixtures is to favour test coverage over churn risk (and in this case the frequency is going to be very rare), and for (b), it would be quite a significant platform change if all apps suddenly gained sensitive secrets in their env without warning or needing to add a addon/enable a labs etc (which would require much notification to customers, let alone deliberation internally), plus the Hatchet test accounts are standard non-privileged test users that have no additional access.

I did try out a loose regex based inclusion list approach:

```
env | sort | grep -E "^(DYNO_RAM|FORWARDED_ALLOW_IPS|GUNICORN.*|LANG|.*PATH.*|PIP.*|PYTHON.*|WEB_CONCURRENCY)="
```

However, as can be see there isn't really a consistent pattern of naming of env vars, so we can never be sure that the test isn't missing a new env var.

vendor/WEB_CONCURRENCY.sh
edmorley: I've decided to add an exit code check for the `nproc` call in daac379d63d493052600be2e152ccafab138c555, to cover some unknown future case where nproc can't calculate the number of available CPUs (perhaps because of some platform misconfiguration), which has the added bonus side-effect of covering the "nproc missing since the buildpack is being run on an unsupported base image" case too :-)

Malax (approved):

edmorley:
Updated the defaults listed at:
https://devcenter.heroku.com/articles/python-concurrency#default-settings-and-behavior

Published changelog entry:
https://devcenter.heroku.com/changelog-items/2846

PR #1569, opened 2024-04-18
Fix the "Python security update available" version check

edmorley:
Previously the "Python security update is available" message was shown if the requested Python version does not exactly equal what the buildpack believes to be the latest Python patch version for that major release.

However, this means that the message is then incorrectly shown if the current version is actually newer than the latest version the buildpack knows about.

This scenario can occur:
(a) During the small time window when a new Python version has been built and uploaded to S3 but the buildpack updates have not yet been released.
(b) If an app pins to an older buildpack version but manually requests a newer Python patch release.

It's not possible to add an integration test for this specific case, since using a fake future version (like Python 3.12.999) will fail prior to the version check due to it not existing on S3. (The scenario being fixed is effectively a race condition that we can't emulate.)

However, the security version numbers are tested in general:
https://github.com/heroku/heroku-buildpack-python/blob/main/spec/hatchet/python_update_warning_spec.rb

GUS-W-15541305.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Fixed the "Python security update is available" warning being shown when the requested version is newer than the latest version known to the buildpack. ([#1569](https://github.com/heroku/heroku-buildpack-python/pull/1569))
 - Fixed glibc warnings seen when downgrading the stack version. ([#1568](https://github.com/heroku/heroku-buildpack-python/pull/1568))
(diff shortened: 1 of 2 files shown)


edmorley:
> Just as a general note, I think this check will fail if we ever decide to throw alpha, beta, RC versions into the mix.

We don't support these for Python. And if we ever added support, we'd add a test, which would catch this. (Also in the event we ever add support in the future, it will likely be for the CNB only, which will have switched to the manifest by then, and also won't be using a bash implementation for version comparison.)

dzuelke (approved):

PR #1570, opened 2024-04-18
Improve the EOL Python error messaging

edmorley:
For cases where a requested Python version is both (a) EOL, and (b) was never built for that stack (such as is the case when we add new stacks), previously the generic "version isn't available for this stack" error message was shown instead of the more specific EOL Python version error message.

Now, the EOL version check is performed first before the S3 presence check, so the more specific EOL message is shown for this case.

In addition to improving the UX, making this change now reduces the test fixture churn both when we add a new stack and for #1567.

I've also dropped the "PyPy is no longer supported" error message and associated test, since very few apps ever used it and it's now been 19 months since support was removed in #1364, so it's fine to show the generic "Python version isn't available" error message for it instead.

GUS-W-15541279.


dzuelke (approved):
Had to look up the `extglob` pattern repetition syntax there ;)

PR #1574, opened 2024-04-22
Add Python binary build support for Heroku-24

edmorley:
This adds support for building Python binaries for Heroku-24 using GitHub Actions. Since the Heroku-24 base image supports both `amd64` and `arm64` (to allow for multi-arch CNBs), for Heroku-24 (only) we now also build binaries for both architectures.

A later PR will add support for Heroku-24 to the buildpack itself, however, that depends on (a) the Python binaries already having been built, (b) the Heroku build system for Heroku-24 being ready, so those changes are not being made yet.

GUS-W-15571102.

### DIFF
--- a/.github/workflows/build_python_runtime.yml
+++ b/.github/workflows/build_python_runtime.yml
@@ -15,6 +15,7 @@ on:
           - auto
           - heroku-20
           - heroku-22
+          - heroku-24
         default: auto
         required: false
       dry_run:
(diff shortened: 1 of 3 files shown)


runesoerensen (approved):
🎉🥳

edmorley:
Builds generated using:

```bash
for v in 3.12.{0..3}; do
  gh workflow run build_python_runtime.yml -F "python_version=${v}" -F "stack=heroku-24"
done
```

(I'll re-run these once Ubuntu 24.04 is GA and the base images refreshed with the final libs versions)

PR #1575, opened 2024-04-22
Add buildpack support for Heroku-24

edmorley:
Following on from #1574 (which added support for building and releasing Python binaries for Heroku-24), this adds support for the new stack to the buildpack itself.

Heroku-24's base images are published as both `amd64` and `arm64` variants (primarily for CNB multi-arch support), so the buildpack now checks the architecture where necessary to ensure it continues to operate should it be run against the `arm64` image outside of Heroku. (However, when run on Heroku, this buildpack is still only ever run on `amd64`.)

GUS-W-14667590.

### DIFF
--- a/.github/workflows/build_python_runtime.yml
+++ b/.github/workflows/build_python_runtime.yml
@@ -70,8 +70,8 @@ jobs:
         run: aws s3 sync ./upload "s3://${S3_BUCKET}"
 
   heroku-24:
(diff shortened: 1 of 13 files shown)


edmorley:
Leaving as draft for now, since CI won't pass until the build system instances for Heroku-24 are ready.

dzuelke (approved):

PR #1579, opened 2024-05-03
Build Python 3.10 runtimes for Heroku-24 too

edmorley:
Expands the Heroku-24 Python version support to include Python 3.10, in addition to the existing Python 3.11 and 3.12.

This will give more Python version options to CNB users that need ARM64 support, given that it will only available be on Heroku-24+.

I've not added a changelog entry since Heroku-24 is not yet released, so no one is using it anyway, and this is covered by the existing Heroku-24 changelog entry (the previous release of the buildpack will be able to use these binaries from S3 too).

GUS-W-15571102.

### DIFF
--- a/.github/workflows/build_python_runtime.yml
+++ b/.github/workflows/build_python_runtime.yml
@@ -70,21 +70,21 @@ jobs:
         run: aws s3 sync ./upload "s3://${S3_BUCKET}"
 
   heroku-24:
-    # On Heroku-24 we only support Python 3.11+.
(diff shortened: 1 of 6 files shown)


edmorley:
Builds triggered using:

```bash
for v in 3.10.{0..14}; do
  gh workflow run build_python_runtime.yml -F "python_version=${v}" -F "stack=heroku-24" --ref h24-python-3.10
done
```

dzuelke (approved):

PR #1581, opened 2024-05-09
Add pre-upload Python runtime tests

edmorley:
We already test the Python runtimes via the buildpack's own tests ([CNB example](https://github.com/heroku/buildpacks-python/blob/de378063c83e1eec6302e79f93f189162b5bbbbf/tests/python_version_test.rs#L121-L123)), however, these are run once the compiled Python runtime has already been uploaded to S3.

This adds tests as part of the compile/package/upload workflow itself.

In addition, these tests now also ensure that all of the optional stdlib modules have been built - since otherwise the upstream CPython configure scripts will happily skip modules where libraries/headers are missing, with only a small warning printed to the logs. These are similar to (but more comprehensive than):
https://github.com/docker-library/official-images/blob/master/test/tests/python-imports/container.py

An example of the new tests showing an optional module hasn't been compiled:
https://github.com/heroku/heroku-buildpack-python/actions/runs/9019650675/job/24782995243#step:6:1862

GUS-W-15727363.

### DIFF
--- a/.github/workflows/build_python_runtime.yml
+++ b/.github/workflows/build_python_runtime.yml
@@ -46,8 +46,13 @@ jobs:
         uses: actions/checkout@v4
       - name: Build Docker image
         run: docker build --platform="linux/amd64" --pull --tag buildenv --build-arg=STACK_VERSION builds/
-      - name: Build and package Python runtime
+      - name: Compile and package Python runtime
         run: docker run --rm --volume="${PWD}/upload:/tmp/upload" buildenv ./build_python_runtime.sh "${{ inputs.python_version }}"
+      - name: Test Python runtime
+        run: |
+          RUN_IMAGE='heroku/heroku:${{ env.STACK_VERSION }}'
+          ARCHIVE_FILENAME='python-${{ inputs.python_version }}-ubuntu-${{ env.STACK_VERSION }}.04-amd64.tar.zst'
+          docker run --rm --volume="${PWD}/upload:/upload:ro" --volume="${PWD}/builds:/builds:ro" "${RUN_IMAGE}" /builds/test_python_runtime.sh "/upload/${ARCHIVE_FILENAME}"
       - name: Upload Python runtime archive to S3
         if: (!inputs.dry_run)
         run: aws s3 sync ./upload "s3://${S3_BUCKET}"
@@ -63,8 +68,13 @@ jobs:
         uses: actions/checkout@v4
       - name: Build Docker image
         run: docker build --platform="linux/amd64" --pull --tag buildenv --build-arg=STACK_VERSION builds/
-      - name: Build and package Python runtime
+      - name: Compile and package Python runtime
         run: docker run --rm --volume="${PWD}/upload:/tmp/upload" buildenv ./build_python_runtime.sh "${{ inputs.python_version }}"
+      - name: Test Python runtime
+        run: |
+          RUN_IMAGE='heroku/heroku:${{ env.STACK_VERSION }}'
+          ARCHIVE_FILENAME='python-${{ inputs.python_version }}-ubuntu-${{ env.STACK_VERSION }}.04-amd64.tar.zst'
+          docker run --rm --volume="${PWD}/upload:/upload:ro" --volume="${PWD}/builds:/builds:ro" "${RUN_IMAGE}" /builds/test_python_runtime.sh "/upload/${ARCHIVE_FILENAME}"
       - name: Upload Python runtime archive to S3
         if: (!inputs.dry_run)
         run: aws s3 sync ./upload "s3://${S3_BUCKET}"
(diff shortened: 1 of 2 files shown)


builds/test_python_runtime.sh
```diff
@@ -0,0 +1,45 @@
...
+# depend on it). Since the Python binary was built in shared mode, `LD_LIBRARY_PATH` must be set
+# when relocating, so the Python binary (which itself contains very little) can find `libpython`.
+INSTALL_DIR=$(mktemp -d)
+PYTHON="${INSTALL_DIR}/bin/python"
+export LD_LIBRARY_PATH="${INSTALL_DIR}/lib/"
+
+tar --zstd --extract --verbose --file "${ARCHIVE_FILEPATH}" --directory "${INSTALL_DIR}"
+
+# Check Python exists at the `python` alias (and not just `python3`) and can run.
+"${PYTHON}" --version
```
schneems, line 18: [Talking out loud - eventual suggestion] If I'm reading this correctly it will expand to `"<tmpdir>/bin/python" --version` which seems to disagree with the comment which I would believe would be checking for `python -v` with that on the path? 

The mention of `python3` seems to be a red herring as it doesn't seem we're executing a named python version i.e. `python3` anywhere. 

So I guess this is REALLY checking that `"<tmpdir>/bin" has a `python` executable in it that responds to `--version` so if that bin was put on PATH it would function correctly. That checks out. 

We are exercising the python binary later in the loop. I think that what you're actually doing here is failing fast if there's a problem with either: 

- Executable permissions
- Expected file does not exist (what your comment says)
- Some problem that results in a working binary that can be executed but does not function correctly, i.e., `--version` fails.

[optional suggestion]: Perhaps we could say that (or some variation):

```suggestion
# Fail if python cannot print the version
"${PYTHON}" --version
```

Which is really already covered by the call itself, but does prevent someone from saying "why is this do-nothing line in here?" so I think it's good to leave some "here's what we're expecting to test with this code.

builds/test_python_runtime.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# Check that all dynamically linked libraries exist in the run image (since it has fewer packages than the build image).
+if find "${INSTALL_DIR}" -name '*.so' -exec ldd '{}' + | grep 'not found'; then
+  echo "The above dynamically linked libraries were not found!"
+  exit 1
+fi
```
schneems, line 24: [Talking out loud] This is neat. You're iterating through all shared object files in the python directory and making sure they're findable via `ldd`. Smart, and produces a good error message stating what expected package was missing. I like it.

.github/workflows/build_python_runtime.yml
```diff
@@ -46,8 +46,13 @@ jobs:
...
       - name: Build Docker image
         run: docker build --platform="linux/amd64" --pull --tag buildenv --build-arg=STACK_VERSION builds/
-      - name: Build and package Python runtime
+      - name: Compile and package Python runtime
         run: docker run --rm --volume="${PWD}/upload:/tmp/upload" buildenv ./build_python_runtime.sh "${{ inputs.python_version }}"
+      - name: Test Python runtime
+        run: |
+          RUN_IMAGE='heroku/heroku:${{ env.STACK_VERSION }}'
+          ARCHIVE_FILENAME='python-${{ inputs.python_version }}-ubuntu-${{ env.STACK_VERSION }}.04-amd64.tar.zst'
+          docker run --rm --volume="${PWD}/upload:/upload:ro" --volume="${PWD}/builds:/builds:ro" "${RUN_IMAGE}" /builds/test_python_runtime.sh "/upload/${ARCHIVE_FILENAME}"
```
schneems, line 55: [talking out loud] I was curious about the `ro` from the docs https://docs.docker.com/storage/volumes/

> -v or --volume: Consists of three fields, separated by colon characters (:). The fields must be in the correct order, and the meaning of each field isn't immediately obvious.

> In the case of named volumes, the first field is the name of the volume, and is unique on a given host machine. For anonymous volumes, the first field is omitted.

> The second field is the path where the file or directory are mounted in the container.

> The third field is optional, and is a comma-separated list of options, such as ro. These options are discussed below.

then later

> The readonly option, if present, causes the bind mount to be [mounted into the container as read-only](https://docs.docker.com/storage/volumes/#use-a-read-only-volume). Can be specified as readonly or to.

Seems good.

schneems (approved):
This is a great idea. I left comments while reviewing, kind of like how GPT will explain it's thinking process. So you get a sense of the kinds of things I looked at and my conclusions.

Overall: One very minor comment change suggestion, it's optional though.

builds/test_python_runtime.sh
edmorley: No, this is checking (a) that python can successfully launch (ie it can be relocated, or that the build didn't fail catastrophically etc), (b) that we set up the `python` symlink too, here:
https://github.com/heroku/heroku-buildpack-python/blob/c233dbd97bda3ff22157561b407231d3a752818b/builds/build_python_runtime.sh#L219-L223

This is similar to the check performed during the CNB integration tests:
https://github.com/heroku/buildpacks-python/blob/de378063c83e1eec6302e79f93f189162b5bbbbf/tests/python_version_test.rs#L121-L123

I'll reword "alias" to "symlink" in the comment to make that clearer.

builds/test_python_runtime.sh
edmorley: Yeah this is borrowed from:
https://github.com/heroku/buildpacks-python/blob/de378063c83e1eec6302e79f93f189162b5bbbbf/tests/python_version_test.rs#L121-L123

PR #1596, opened 2024-06-13
Improve shellcheck coverage

edmorley:
* Switches to using `git ls-files` instead of a hardcoded list of scripts to lint (shellcheck doesn't support recursively scanning for scripts itself). This means a few files that were not previously linted now are, so need fixes.
* Adds a `.shellcheckrc` which is used to enable optional shellcheck checks via `enable=all`. Most of these checks either already passed or have been fixed in this PR - however, a few others need a closer look so have been disabled for now to reduce the size of this PR.
* As part of doing this I discovered an NLTK bug with setting `NLTK_DATA` after #1595, which has now been fixed. Since that change was not yet released, I've not mentioned the fix here in the changelog.

Shellcheck's optional checks as of v0.10.0:

```
$ shellcheck --list-optional
name:    add-default-case
desc:    Suggest adding a default case in `case` statements
example: case $? in 0) echo 'Success';; esac
fix:     case $? in 0) echo 'Success';; *) echo 'Fail' ;; esac

name:    avoid-nullary-conditions
desc:    Suggest explicitly using -n in `[ $var ]`
example: [ "$var" ]
fix:     [ -n "$var" ]

name:    check-extra-masked-returns
desc:    Check for additional cases where exit codes are masked
example: rm -r "$(get_chroot_dir)/home"
fix:     set -e; dir="$(get_chroot_dir)"; rm -r "$dir/home"

name:    check-set-e-suppressed
desc:    Notify when set -e is suppressed during function invocation
example: set -e; func() { cp *.txt ~/backup; rm *.txt; }; func && echo ok
fix:     set -e; func() { cp *.txt ~/backup; rm *.txt; }; func; echo ok

name:    check-unassigned-uppercase
desc:    Warn when uppercase variables are unassigned
example: echo $VAR
fix:     VAR=hello; echo $VAR

name:    deprecate-which
desc:    Suggest 'command -v' instead of 'which'
example: which javac
fix:     command -v javac

name:    quote-safe-variables
desc:    Suggest quoting variables without metacharacters
example: var=hello; echo $var
fix:     var=hello; echo "$var"

name:    require-double-brackets
desc:    Require [[ and warn about [ in Bash/Ksh
example: [ -e /etc/issue ]
fix:     [[ -e /etc/issue ]]

name:    require-variable-braces
desc:    Suggest putting braces around all variable references
example: var=hello; echo $var
fix:     var=hello; echo ${var}
```

See also:
https://github.com/koalaman/shellcheck/blob/master/shellcheck.1.md

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,15 +9,8 @@ STACK_IMAGE_TAG := heroku/$(subst -,:,$(STACK))-build
 
 lint: lint-scripts lint-ruby
 
-# TODO: Enable scanning for files that are currently missed and/or restructure repo
-# layout to make it more viable to use wildcards here, given:
-# https://github.com/koalaman/shellcheck/issues/962
 lint-scripts:
-	@shellcheck -x bin/compile bin/detect bin/release bin/test-compile bin/utils bin/warnings bin/default_pythons
-	@shellcheck -x bin/steps/collectstatic bin/steps/nltk bin/steps/pip-install bin/steps/pipenv bin/steps/pipenv-python-version bin/steps/python
-	@shellcheck -x bin/steps/hooks/*
-	@shellcheck -x vendor/WEB_CONCURRENCY.sh
-	@shellcheck -x builds/*.sh
+	@git ls-files --cached --others --exclude-standard 'bin/*' '*/bin/*' '*.sh' | xargs shellcheck --check-sourced --color=always
 
 lint-ruby:
 	@bundle exec rubocop
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ STACK_IMAGE_TAG := heroku/$(subst -,:,$(STACK))-build
 lint: lint-scripts lint-ruby
 
 lint-scripts:
-	@git ls-files --cached --others --exclude-standard 'bin/*' '*/bin/*' '*.sh' | xargs shellcheck --check-sourced --color=always
+	@git ls-files -z --cached --others --exclude-standard 'bin/*' '*/bin/*' '*.sh' | xargs -0 shellcheck --check-sourced --color=always
 
 lint-ruby:
 	@bundle exec rubocop
--- a/bin/steps/collectstatic
+++ b/bin/steps/collectstatic
@@ -10,7 +10,9 @@
 #   - $DISABLE_COLLECTSTATIC: disables this functionality.
 #   - $DEBUG_COLLECTSTATIC: upon failure, print out environment variables.
 
-# This script is run in a subshell via sub_env so doesn't inherit the vars/utils from `bin/compile`.
+# This script is run in a subshell via sub_env so doesn't inherit the options/vars/utils from `bin/compile`.
+# TODO: Stop running this script in a subshell.
+set -eo pipefail
 BUILDPACK_DIR=$(cd "$(dirname "$(dirname "$(dirname "${BASH_SOURCE[0]}")")")" && pwd)
 source "${BUILDPACK_DIR}/bin/utils"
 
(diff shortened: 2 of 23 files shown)


Makefile
```diff
@@ -9,15 +9,8 @@ STACK_IMAGE_TAG := heroku/$(subst -,:,$(STACK))-build
...
-# TODO: Enable scanning for files that are currently missed and/or restructure repo
-# layout to make it more viable to use wildcards here, given:
-# https://github.com/koalaman/shellcheck/issues/962
 lint-scripts:
-	@shellcheck -x bin/compile bin/detect bin/release bin/test-compile bin/utils bin/warnings bin/default_pythons
-	@shellcheck -x bin/steps/collectstatic bin/steps/nltk bin/steps/pip-install bin/steps/pipenv bin/steps/pipenv-python-version bin/steps/python
-	@shellcheck -x bin/steps/hooks/*
-	@shellcheck -x vendor/WEB_CONCURRENCY.sh
-	@shellcheck -x builds/*.sh
+	@git ls-files --cached --others --exclude-standard 'bin/*' '*/bin/*' '*.sh' | xargs shellcheck --check-sourced --color=always
```
dzuelke, line 13:
```suggestion
	@git ls-files -z --cached --others --exclude-standard 'bin/*' '*/bin/*' '*.sh' | xargs -0 shellcheck --check-sourced --color=always
```

bin/steps/collectstatic
```diff
@@ -10,7 +10,9 @@
 #   - $DISABLE_COLLECTSTATIC: disables this functionality.
 #   - $DEBUG_COLLECTSTATIC: upon failure, print out environment variables.
 
-# This script is run in a subshell via sub_env so doesn't inherit the vars/utils from `bin/compile`.
+# This script is run in a subshell via sub_env so doesn't inherit the options/vars/utils from `bin/compile`.
+# TODO: Stop running this script in a subshell.
+set -eo pipefail
```
dzuelke, line 15:
```suggestion
set -euo pipefail
```

(right?)

bin/steps/python
```diff
@@ -102,8 +105,8 @@ if [[ "$STACK" != "$CACHED_PYTHON_STACK" ]]; then
     rm -rf .heroku/python-stack .heroku/python-version .heroku/python .heroku/vendor .heroku/python .heroku/python-sqlite3-version
 fi
 
-if [ -f .heroku/python-version ]; then
-  if [ ! "$(cat .heroku/python-version)" = "$PYTHON_VERSION" ]; then
+if [[ -f .heroku/python-version ]]; then
+  if [[ ! "$(cat .heroku/python-version)" = "$PYTHON_VERSION" ]]; then
```
dzuelke, line 109:
```suggestion
  if [[ ! "$(cat .heroku/python-version)" == "$PYTHON_VERSION" ]]; then
```

bin/utils
```diff
@@ -7,7 +7,7 @@ shopt -s nullglob
 
 source "${BUILDPACK_DIR}/vendor/buildpack-stdlib_v8.sh"
 
-if [ "$(uname)" == Darwin ]; then
+if [[ "$(uname)" == Darwin ]]; then
```
dzuelke, line 10:
```suggestion
if [[ "$(uname)" == "Darwin" ]]; then
```

(quote both or neither)

dzuelke (approved):

bin/steps/collectstatic
edmorley: Eventually, but it breaks things if I enable it now. I was going to enable `-u` at the same time as doing to for `bin/compile`. There are just too many things to fix/change in one PR, so I'm splitting it up.

Makefile
edmorley: Given that ls-files quotes any unusual filenames, why is this needed?

dzuelke: It doesn't quote for correct shell/xargs consumption:

```ShellSession
$ git ls-files foo*
"foo\nbar"
foo bar
$ git ls-files foo* | xargs cat
cat: foo\nbar: No such file or directory
cat: foo: No such file or directory
cat: bar: No such file or directory
```

PR #1597, opened 2024-06-17
Implement `bin/report`

edmorley:
Implements the `bin/report` build report API, as a replacement for log based metrics (which stopped working some time ago with the change in internal observability vendor).

This will help with varies upcoming Python buildpack changes, such as the EOL of Python 3.8, assessing usage levels of different package managers (wrt Poetry addition, and future of Pipenv for CNB), removal of the legacy sqlite feature etc.

This implementation is based on that used by the Node.js buildpack, which is currently the only buildpack that implements `bin/report`. There are a number of rough edges with the Node.js implementation that in an ideal world we'd improve upon. However, given that that implementation works, is actively used in production, and CNBs are our priority moving forwards, I've chosen not to make many changes to that design.

See:
https://github.com/heroku/heroku-buildpack-nodejs/blob/main/bin/report https://github.com/heroku/heroku-buildpack-nodejs/blob/main/lib/metadata.sh https://github.com/heroku/heroku-buildpack-nodejs/blob/main/lib/kvstore.sh

GUS-W-8047826.


bin/steps/sqlite3
```diff
@@ -64,11 +64,16 @@ sqlite3_install() {
...
   puts-step "Installing SQLite3"
 
+  # TODO: This never actual prints failure or even aborts the build, since
+  # the conditional disables `set -e` inside the called function:
+  # https://stackoverflow.com/q/19789102
+  # ...plus whoever wrote this forgot the `exit 1` in the `else` anyway.
   if sqlite3_install "$BUILD_DIR/.heroku/python" ; then
-    mcount "success.python.sqlite3"
+    # mcount "success.python.sqlite3"
+    true
```
dzuelke, line 73:
```suggestion
    :
```

See https://man7.org/linux/man-pages/man1/colon.1p.html or `help ":"`:

> ```
> bash-5.2$ help ":"
> :: :
>     Null command.
>     
>     No effect; the command does nothing.
>     
>     Exit Status:
>     Always succeeds.
> ```

dzuelke (approved):

PR #1599, opened 2024-06-20
Support cgroups v1 and v2 for memory limits

dzuelke:
For v2, it uses `memory.high` if available (recommended way of setting a functioning soft limit), then first falls back to `memory.max` (e.g. from `docker run -m`), then to `memory.low` (e.g. from `docker run --memory-reservation`), and finally to `memory.min`.

Falls back to direct reading of `/sys/fs/cgroup/memory/memory.limit_in_bytes` for cases where that exists, but no full cgroupfs is mounted (e.g. on Heroku).

Limit enforcement (now to 8 TB) is still in place this way - a Docker v1 value will be read, and run into the limit for unrestricted containers, without hitting the fallback and getting returned.

GUS-W-16051944

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Support reading container memory limits from cgroups (v1 and v2)
 
 ## [v252] - 2024-06-17
 
--- a/bin/compile
+++ b/bin/compile
@@ -271,8 +271,11 @@ EOT
 fi
 
 # Install sane-default script for $WEB_CONCURRENCY and $FORWARDED_ALLOW_IPS.
-cp "${BUILDPACK_DIR}/vendor/WEB_CONCURRENCY.sh" "$WEB_CONCURRENCY_PROFILE_PATH"
 cp "${BUILDPACK_DIR}/vendor/python.gunicorn.sh" "$GUNICORN_PROFILE_PATH"
+# concatenate these together to avoid possible collisions with other buildpacks that use the same cgroup utils under the same name
+# in case of differences, the last buildpack that ran would "win"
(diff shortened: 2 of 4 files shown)


edmorley (changes requested):
Summary from our call yesterday:
- Overall complexity/size of implementation is significantly larger than what we had been expecting based on the initial proposal. If this were being designed for a standalone open source library for handling all versions of cgroups then fair enough - but that wasn't the remit here (which was "add cgroups v2 support in a minimal way to our classic buildpacks").
- The CLI arg parsing should be removed - some options are unused, others are for things that don't need to be configurable since all langs can agree on a sensible default (eg we can all use the same threshold for the failsafe "is this the bogus petabytes memory value or not" check), others are for testing upstream but that could be handled via eg env vars or similar instead.
- This PR currently leaves behind the plain "v1 memory limits file check" in `WEB_CONCURRENCY.sh` - this should be moved into the cgroups functions so they handle everything.
- We don't need/want full cgroups v1 support, since it can't fully replace the hardcoded `/sys/fs/cgroup/memory/memory.limit_in_bytes` file check anyway (since we need to support the case where the file has been faked but cgroups v1 isn't enabled officially), at which point there is little benefit to including the cgroups v1 implementation too.
- Whilst I can understand why the implementation has been split into so many functions, I feel like in some cases it's hurt readability slightly (eg it's much harder to follow the flow linearly here, particularly since so many functions have similar names with terms that don't mean much to people who aren't intimately familiar with the cgroups spec), so maybe inlining might be better in a couple of places (not all)?

edmorley:
Cancelling review pending further discussion as a team as to next steps wrt this project.

edmorley:
Closing this out until we decide the plan here.

PR #1608, opened 2024-06-28
Improve detection error messages

edmorley:
Previously if an app was missing the necessary Python package manager files, buildpack detection would fail with the generic error message provided by the build system (see previous version of test fixtures for an example).

This error message required users to click through to the Dev Center documentation to find out what files need to be present for a Python app to work on Heroku, and there are quite a few support tickets where users have struggled to resolve this themselves (see GUS work item for ticket links).

Now a custom error message is displayed, similar to what the Node.js and PHP buildpacks do:
https://github.com/heroku/heroku-buildpack-nodejs/blob/main/bin/detect
https://github.com/heroku/heroku-buildpack-php/blob/main/bin/detect

This error message lists some common reasons why the files might not be found (such as them not being in the root directory), and also prints a directory listing of the root directory to help the user (and support) understand what is happening.

Lastly, the detection logic has been made more permissive, so that Python apps that are missing some of the required files (but have other misc Python related files) still pass detection, allowing us to perform the stricter file check during the build phase instead. This is particularly important for the "new app with no buildpacks set" case, where in the case of buildpack detect failure a custom detect error message is not shown by the build system (since it's not viable to show the custom message from half a dozen language buildpacks simultaneously), but a build phase error message is.

These changes also bring the classic Python buildpack more in line with the Python CNB implementation:
https://github.com/heroku/buildpacks-python/blob/447c7dd978bc3069fb1baa0005a83e4720ba1e63/src/detect.rs#L4-L25
https://github.com/heroku/buildpacks-python/blob/447c7dd978bc3069fb1baa0005a83e4720ba1e63/src/errors.rs#L71-L81

GUS-W-7924407.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -134,6 +134,41 @@ else
   CACHED_PYTHON_STACK=$STACK
 fi
 
+# TODO: Move this into a new package manager handling implementation when adding Poetry support.
+if [[ ! -f requirements.txt && ! -f Pipfile && ! -f setup.py ]]; then
+  puts-warn
+  puts-warn "Error: No supported Python package manager files were found."
+  puts-warn
+  # We intentionally don't mention `setup.py` here since it's being removed soon.
+  puts-warn "A Python app on Heroku must have either a 'requirements.txt' or"
+  puts-warn "'Pipfile' file in the root directory of its source code, so the"
+  puts-warn "buildpack knows which dependencies to install."
+  puts-warn
+  puts-warn "Currently the root directory of your app contains:"
+  puts-warn
+  # TODO: Overhaul logging helpers so they can handle prefixing multi-line strings, and switch to them.
+  # shellcheck disable=SC2012 # Using `ls` instead of `find` is absolutely fine for this use case.
+  ls -1 --indicator-style=slash "${BUILD_DIR}" | sed 's/^/ !     /'
+  puts-warn
+  puts-warn "If you believe your app already has a 'requirements.txt' or"
+  puts-warn "'Pipfile' file, check that:"
+  puts-warn
+  puts-warn "1. The file is in the top level directory (not a subdirectory)."
+  puts-warn "2. The filename has the correct spelling and capitalisation."
+  puts-warn "3. The filename isn't listed in '.gitignore' or '.slugignore'."
+  puts-warn
+  puts-warn "Otherwise, please create a 'requirements.txt' file in the root"
+  puts-warn "of your app source, which lists your app's Python dependencies"
+  puts-warn "(the file can be empty if your app has no dependencies)."
+  puts-warn
+  puts-warn "For help with using Python on Heroku, see:"
+  puts-warn "https://devcenter.heroku.com/articles/getting-started-with-python"
+  puts-warn "https://devcenter.heroku.com/articles/python-support"
+  puts-warn
+  meta_set "failure_reason" "package-manager-not-found"
+  exit 1
+fi
+
 # Pipenv Python version support.
 # Detect the version of Python requested from a Pipfile (e.g. python_version or python_full_version).
 # Convert it to a runtime.txt file.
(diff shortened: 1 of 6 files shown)


bin/compile
```diff
@@ -134,6 +134,41 @@ else
   CACHED_PYTHON_STACK=$STACK
 fi
 
+# TODO: Move this into a new package manager handling implementation when adding Poetry support.
+if [[ ! -f requirements.txt && ! -f Pipfile && ! -f setup.py ]]; then
+  puts-warn
+  puts-warn "Error: No supported Python package manager files were found."
+  puts-warn
+  # We intentionally don't mention `setup.py` here since it's being removed soon.
```
colincasey, line 142: How soon is soon? If we're talking longer than a day or two, maybe you should mention it until it truly is removed?

bin/compile
```diff
@@ -134,6 +134,41 @@ else
...
+  puts-warn
+  # TODO: Overhaul logging helpers so they can handle prefixing multi-line strings, and switch to them.
+  # shellcheck disable=SC2012 # Using `ls` instead of `find` is absolutely fine for this use case.
+  ls -1 --indicator-style=slash "${BUILD_DIR}" | sed 's/^/ !     /'
+  puts-warn
+  puts-warn "If you believe your app already has a 'requirements.txt' or"
+  puts-warn "'Pipfile' file, check that:"
+  puts-warn
+  puts-warn "1. The file is in the top level directory (not a subdirectory)."
+  puts-warn "2. The filename has the correct spelling and capitalisation."
```
colincasey, line 157: Not really an issue but something I'm curious about - do we prefer `en-US` spelling over other locales where this would be `capitalization` instead of `capitalisation`?  I doubt most folks would notice this but, as someone who's done localization work and as a fellow non-American, inconsistencies like this stand out to me 😅

joshwlewis (approved):

bin/compile
```diff
@@ -134,6 +134,41 @@ else
...
+if [[ ! -f requirements.txt && ! -f Pipfile && ! -f setup.py ]]; then
+  puts-warn
+  puts-warn "Error: No supported Python package manager files were found."
+  puts-warn
+  # We intentionally don't mention `setup.py` here since it's being removed soon.
+  puts-warn "A Python app on Heroku must have either a 'requirements.txt' or"
+  puts-warn "'Pipfile' file in the root directory of its source code, so the"
+  puts-warn "buildpack knows which dependencies to install."
+  puts-warn
+  puts-warn "Currently the root directory of your app contains:"
```
joshwlewis, line 147: I think the message would be just as clear without listing the files. And this long error message could get very long in a project with many files. I'd suggest going without the directory listing. (optional)

colincasey (approved):

bin/compile
edmorley: So IMO the directory listing is quite an important part of the error message, since in many of the cases users are adamant (eg in support tickets) that the files are present, but it turns out they were miss-spelt or not in the top level directory.

By listing the files in the error message - it gives the user a greater chance to spot the issue, and if they still don't notice it, it means support can diagnose without needing to ask for a screenshot of file listings or having the user push to a support branch.

Re the listing becoming unwieldy - the error message only lists the top level directory (rather than recursively), and the primary target of these error messages is new projects, which typically won't have hundreds of files in the root directory.

It's worth noting that the directory listing is inspired by what the Node.js buildpack already does:
https://github.com/heroku/heroku-buildpack-nodejs/blob/c9931b23b297c71199e2777144f75d58d6d977a5/bin/detect#L64

bin/compile
edmorley: I'll be removing it in the next week. However, I don't want new projects to start using it, so regardless of timeline I don't want to advertise it. (Only <0.1% of builds use it currently.)

bin/compile
edmorley: At Mozilla en-US vs en-GB was document-author/component-owners call. I don't know about here :-)

Issue #932, opened 2020-03-24
Python Version via .python-version


jamesward:
It doesn't seem like there is a standard way for a project to specify it's Python version, but I'm using https://github.com/pyenv/pyenv and it does this via a `.python-version` file.  Is adding support for reading the version from that file something that would be considered?

CaseyFaist:
Absolutely! I can't promise when work will start on this feature specifically, but support for Poetry and using up to date python conventions is very much on my mind.

joshfriend:
@jamesward have you discovered Heroku's [`runtime.txt`](https://devcenter.heroku.com/articles/python-runtimes#selecting-a-runtime) file yet?

There are community buildpacks for pyenv, but pyenv compiles python from source (lengthening your slug build time) whereas this buildpack has prebuilt python binaries for common versions.

Ideally the `runtime.txt` and `.python-version` files would be treated as interchangeable by this buildpack, since the format and names of the python versions used are _almost_ identical.

jamesward:
I have seen that but it'd be nice to be able to specify the python version in just one place. :)

abitrolly:
[`pyproject.toml`](https://snarky.ca/what-the-heck-is-pyproject-toml/) can also be used to avoid yet another file.

There is a [`requires-python`](https://www.python.org/dev/peps/pep-0621/#requires-python) directive, but it is designed to describe [all Python versions](https://packaging.python.org/specifications/core-metadata/#requires-python) a package is compatible with. For the `buildpack`, however, it may beneficial to set specific version in  `[tool.buildpack]`

```
[tool.buildpack]
python = "3.8.x"
```

edmorley:
I'd love to be able to add support for this in a vendor-independent manner.

Perhaps we can use `requires-python` and either:
(a) insist it specifies a single version (or at least using only a fairly easy to manually parse version specifier),
(b) resolve the version range ourselves (similar to what the Node buildpack does), though we'll then have a pre-python-install-time dependency on `packaging` or whichever package it is

edmorley:
I've written up my latest thoughts on ways to specify Python version for Heroku apps, here:
https://github.com/heroku/heroku-buildpack-python/issues/913#issuecomment-1031468659

edmorley:
Sorry for the delay in getting to this! 

[...]

PR #1640, opened 2024-09-24
Refactor package manager handling

edmorley:
Previously, the buildpack didn't track which package manager was being used, but instead had a number of fragile file presence checks and env vars like `SKIP_PIP_INSTALL` scattered throughout (which won't work when the package manager isn't a binary choice).

As such, package manager handling has been refactored to allow for more easily adding Poetry support later.

This also continues the gradual migration of code from sourced `bin/steps/` scripts, to functions declared in `lib/`. As part of this, I've started namespacing the lib functions using the `module::fn_name()` pattern described in:
https://google.github.io/styleguide/shellguide.html#function-names

On the most part I've tried to keep this refactoring a no-op (aside from fixes for a few quirks/bugs found whilst working on this). Longer term we will be deprecating/sunsetting several features, such as the `setup.py` fallback, using Pipenv without a lockfile, and support for apps that have ended up with the files for multiple package managers.

Prep for #796 / #932.

GUS-W-16811078.


lib/package_manager.sh
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env bash
+
+function package_manager::determine_package_manager() {
+	local build_dir="${1}"
+	local package_managers_found=()
```
colincasey, line 5: Based on the logic below, an indexed array will work fine but I do think you should consider an associative array here instead. Mostly as a way to guarantee you only have one unique entry per package manager detected to prevent a small change in logic introducing a subtle bug where you could unintentionally end up with a double entry (e.g.; `package_managers_found=(pip pip)`)

bin/steps/python
```diff
@@ -166,34 +166,3 @@ else
...
-if [[ -z "${BUNDLED_PIP_WHEEL}" ]]; then
-	display_error "Error: Failed to locate the bundled pip wheel."
-	meta_set "failure_reason" "bundled-pip-not-found"
-	exit 1
-fi
-
-/app/.heroku/python/bin/python "${BUNDLED_PIP_WHEEL}/pip" install --quiet --disable-pip-version-check --no-cache-dir \
-	"pip==${PIP_VERSION}" "setuptools==${SETUPTOOLS_VERSION}" "wheel==${WHEEL_VERSION}"
-
-hash -r
```
colincasey, line 199: I couldn't find where this `hash -r` command moved during this refactor. Is it no longer required after the pip install?

colincasey (approved):
Looks okay to me @edmorley. Just had a couple of minor comments for you to respond to.

lib/package_manager.sh
edmorley: It also seems that with associative arrays:
- it's not possible to fetch a single element (only "display all keys")
- the keys when listed are not listed in insertion order, which would break the package manager order of precedence requirement of the current implementation

PR #1656, opened 2024-10-06
Add support for Python 3.9 on Heroku-24

edmorley:
Previously Python 3.9 was only supported on the Heroku-22 stack and older, since it [reaches EOL in Oct 2025](https://devguide.python.org/versions/#supported-versions).

However, adding support for Python 3.9 to Heroku-24 will make it easier for users upgrading stacks, and also give more Python version options to CNB users that need ARM64 support, given that multi-arch is only available for Heroku-24+. It also simplifies the version error checks we need to perform in the upcoming Python version handling refactor.

GUS-W-16917997.

### DIFF
--- a/.github/workflows/build_python_runtime.yml
+++ b/.github/workflows/build_python_runtime.yml
@@ -80,8 +80,8 @@ jobs:
         run: aws s3 sync ./upload "s3://${S3_BUCKET}"
 
   heroku-24:
-    # On Heroku-24 we only support Python 3.10+.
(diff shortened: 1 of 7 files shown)


edmorley:
Builds triggered using:

```
for v in 3.9.{0..20}; do
  gh workflow run build_python_runtime.yml -F "python_version=${v}" -F "stack=heroku-24" --ref python-3-9-heroku-24
done
```

https://github.com/heroku/heroku-buildpack-python/actions/workflows/build_python_runtime.yml?query=branch%3Apython-3-9-heroku-24

runesoerensen (approved):

Issue #913, opened 2019-12-26
Support flexible patch version syntax to allow for automatic Python updates


abitrolly:
Pinning Python runtime on Heroku prevents it from getting security updates.
```
remote: -----> Python app detected
remote:  !     Python has released a security update! Please consider upgrading to python-3.6.10
remote:        Learn More: https://devcenter.heroku.com/articles/python-runtimes
```
The secure way is to allow any newer patch version to be installed. Proposed syntax.
```
python-3.6.x
```

CaseyFaist:
Heckin yes. Not the next thing on my list, but on the list.

abitrolly:
I could probably help you with a patch given the exact place where Python selection happens.

edmorley:
@abitrolly Hi! Thank you for filing this. I 100% agree this is something we should support. Whilst some projects will always want to specify an exact Python version to ensure determinism/allow control of when the upgrade occurs, there are likely many more that would like to not have to think about patch updates and have them be performed automatically.

Some thoughts that came to mind:
1. What syntax(s) should we support? (eg `python-3.6` vs `python-3.6.x` vs `python-3.6.*`, or even variants without the `python-` prefix)
2. Longer term do we see `runtime.txt` being superceded by something else (eg #932)? If so, does that affect whether we add support for this feature to `runtime.txt` (vs adding only to the new thing), or should we just go ahead with this anyway?

To try and answer these, I took a look at what other tools/services use currently:

### Pyenv
* The `.python-version` file can contain one or more Python versions (separated by newlines)
* Valid version names are those [here](https://github.com/pyenv/pyenv/tree/master/plugins/python-build/share/python-build) - ie: `X.Y.Z` form for CPython (though `python-X.Y.Z` also supported), and `PREFIX-X.Y.X` for other build types.
* No support for flexible patch versions (ref pyenv/pyenv/issues/1349 and pyenv/pyenv/issues/723)

### Python Poetry
* https://python-poetry.org/docs/pyproject/#dependencies-and-dev-dependencies + https://python-poetry.org/docs/dependency-specification/
* Typically doesn't track exact Python version to be used for a project, but more the semver range of supported versions, via a `[tool.poetry.dependencies]` field in `pyproject.toml`
* Example `python = "^3.6"`

### Pipenv
* https://pipenv.pypa.io/en/latest/basics/#specifying-versions-of-python + https://pipenv.pypa.io/en/latest/advanced/#automatic-python-installation
[...]

PR #1658, opened 2024-10-08
Refactor Python version handling

edmorley:
The previous implementation of Python version detection/resolution revolved heavily around the `runtime.txt` file, even though the version could originate from other sources. For example, a fake `runtime.txt` would be written out into the build directory containing the desired Python version, even if the version originated from `Pipfile.lock`.

This meant all later Python version handling in the buildpack would see a `runtime.txt` file, along with versions specified in the syntax of that file (e.g. `python-N.N.N`) strings, even though that wasn't the format in which the user had specified the version.

Now, the buildpack explicitly tracks the requested version and its origin (rather than using the `runtime.txt` file as an API), along with the resolved Python major and full versions (which makes later Python version conditionals less fragile).

In addition, the Python version specifiers are validated upfront at the point of parsing the relevant data source, so that clearer error messages can be shown.

Lastly, the Python version resolution (the mapping of major Python versions to the latest patch release) has been decoupled from the Pipenv version implementation and made more robust, so it can also be used by the upcoming `.python-version` file support.

Prep for #932 / #913.

GUS-W-16821309.
GUS-W-7924371.


schneems (approved):
I read through the warnings/errors and general logic. Read through the test changes. No changes requested.

PR #1661, opened 2024-10-09
Add support for Python 3.13

edmorley:
Add support for Python 3.13, and release 3.13.0.

The default Python version remains unchanged (at 3.12.x) for now.

Notably for Python 3.13 we now:
- No longer install setuptools and wheel - matching what the wider ecosystem has already done for Python 3.12+. (See the Python CNB's removal PR for more details: https://github.com/heroku/buildpacks-python/pull/243)
- No longer install the SQLite headers and CLI, as the first step towards dropping that rarely used feature.
- Have to disable some of the tests that run during the PGO profiling phase when on Heroku-22 (see comments below for more details).

In addition, for all Python versions we now also remove the `idle3` and `pydoc3` scripts, since they do not work with relocated Python and so have been broken for some time. Their functionality continues to be available by invoking them via their modules instead (e.g. `python -m pydoc`).

Release announcement:
https://blog.python.org/2024/10/python-3130-final-released.html https://www.python.org/downloads/release/python-3130/

Details on what's new in Python 3.13:
https://docs.python.org/3.13/whatsnew/3.13.html

Binary builds:
https://github.com/heroku/heroku-buildpack-python/actions/runs/11280580537

Python 3.13 readiness status of the top 360 packages on PyPI: https://pyreadiness.org/3.13/

GUS-W-14846826.
GUS-W-14846839.
GUS-W-16944574.


edmorley:
The binary builds of Python 3.13.0 succeeded on Heroku-20 and Heroku-24, however, failed on Heroku-22 (Ubuntu 22.04) due to failures during the Python tests that are run as part of generating the PGO profile:

```
0:01:48 load avg: 0.93 [43/44] test_xml_etree
test test_xml_etree failed
0:01:49 load avg: 0.93 [44/44] test_xml_etree_c -- test_xml_etree failed (3 failures)
test test_xml_etree_c failed
test_xml_etree_c failed (3 failures)
```

Sadly that is all the output shown. (Build logs at: https://github.com/heroku/heroku-buildpack-python/actions/runs/11259777638/job/31310276989#step:4:1374)

A retrigger of the build didn't resolve the issue.

The offending tests are these:
https://github.com/python/cpython/blob/v3.13.0/Lib/test/test_xml_etree.py
https://github.com/python/cpython/blob/v3.13.0/Lib/test/test_xml_etree_c.py

edmorley:
This might be related to:
https://github.com/python/cpython/issues/125067

edmorley:
I've added a patch that disables the affected tests for now, since:
1. These tests are only used during PGO profile generation, so are just a way that the build process uses to identify hot paths, so the exact tests that run isn't super critical.
2. These tests are failing on Ubuntu 22.04 for older Python too - the only difference now is that the PGO build process no longer ignores the non-zero exit code of the test runner.
3. From the linked CPython issue It looks most likely that the problem is a test issue rather than an issue with Python's xml stdlib functionality itself, so it's safe to just skip the failing tests.

colincasey (approved):

edmorley:
Updated:
https://devcenter.heroku.com/articles/python-runtimes

Posted:
https://devcenter.heroku.com/changelog-items/3002

PR #1664, opened 2024-10-14
Add support for the `.python-version` file

edmorley:
Following on from the recent Python version refactoring in #1658, this now adds support for configuring the app's Python version using a `.python-version` file. This file is used by several tools in the Python ecosystem (such as pyenv, `actions/setup-python`, uv), whereas the existing `runtime.txt` file is proprietary to Heroku.

This change is the classic Python buildpack equivalent of the Python CNB change here:
https://github.com/heroku/buildpacks-python/pull/272

If both a `runtime.txt` file and a `.python-version` file are present, then the `runtime.txt` file will take precedence. However, use of the `.python-version` file is now recommended, since `runtime.txt` will be deprecated in the future. Both the `runtime.txt` file and `.python-version` file take precedence over any Python version specified in a `Pipfile.lock` for Pipenv users.

We support the following `.python-version` syntax:
- Major Python version (e.g. `3.13`, which will then be resolved to the latest Python 3.13). (This form is recommended, since it allows for Python security updates to be pulled in without having to manually bump the version.)
- Exact Python version (e.g. `3.13.0`)
- Comments (lines starting with `#`)
- Blank lines

We don't support the following `.python-version` features:
- Specifying multiple Python versions
- Prefixing versions with `python-` (since this form is undocumented and will likely be deprecated by pyenv in the future)

In addition, the existing `runtime.txt` support has been updated to allow specifying just the major Python version, in order to increase feature parity between the files, and avoid confusion if users try to use the major version only syntax from `.python-version` in their `runtime.txt`.

Refs #913.
Refs #932.
GUS-W-7671453.
GUS-W-16821357.


edmorley:
Created entry in the public roadmap:
https://github.com/heroku/roadmap/issues/317

PR #1682, opened 2024-10-31
Add support for Poetry

### LINKED ISSUE #796: Support for Poetry

edmorley:
After [many](https://github.com/heroku/heroku-buildpack-python/compare/d13fbbdea7a86471872241f811eaf08787757d44...4df96c285ed1aeed4ecf6909887165d4fc712381) refactoring/preparation PRs, we're now ready to add support for the package manager Poetry:
https://python-poetry.org

To use Poetry, apps must have a `poetry.lock` lockfile, which can be created by running `poetry lock` locally, after adding Poetry config to `pyproject.toml` (which can be done either manually or by using `poetry init`).

For now, if a `requirements.txt` or `Pipfile` is found it will take precedence over `poetry.lock` for backwards compatibility (in the future this will become a warning then an error). This means users of the third-party [python-poetry-buildpack](https://github.com/moneymeets/python-poetry-buildpack) will need to remove that buildpack in order to use the new native Poetry support, since it exports a `requirements.txt` file during the build.

Poetry is installed into the build cache rather than the slug, so is not available at run-time (since it's not typically needed at run-time and doing so reduces the slug size). The entrypoints of installed dependencies are available on `PATH`, so use of `poetry run` or `poetry shell` is not required at run-time to use dependencies in the environment.

When using Poetry, pip is not installed since Poetry includes its own internal vendored copy that it will use instead (for the small number of Poetry operations for which it still calls out to pip, such as package uninstalls).

During normal (non-CI) builds, the `poetry install --sync` command is run using `--only main` so as to only install the main `[tool.poetry.dependencies]` dependencies group from `pyproject.toml` and not any of the app's other dependency groups (such as test/dev/... groups, eg `[tool.poetry.group.test.dependencies]`).

On Heroku CI, all default Poetry dependency groups are installed (i.e. all groups minus those marked as `optional = true`).

Relevant Poetry docs:
- https://python-poetry.org/docs/cli/#install
- https://python-poetry.org/docs/configuration/
- https://python-poetry.org/docs/managing-dependencies/#dependency-groups

See also the Python CNB equivalent of this PR:
- https://github.com/heroku/buildpacks-python/pull/261

Note: We don't support controlling the Python version via Poetry's `tool.poetry.dependencies.python` field, since that field typically contains a version range, which is not safe to use. Use the newly added `.python-version` file support instead. For more on this, see the longer explanation over in the Python CNB repo:
https://github.com/heroku/buildpacks-python/issues/260

Closes #796.
Closes #835.
GUS-W-16810914.


runesoerensen (approved):

edmorley:
https://devcenter.heroku.com/changelog-items/3050

PR #1690, opened 2024-11-08
Improve NLTK downloader log output and error messages

edmorley:
Improves the build log output, error messages and buildpack failure metrics for the NLTK downloader feature, and adds a test for the app having an invalid `nltk.txt` (since previously only the successful cases were tested).

This should be the last failure type seen in the wild that doesn't emit a `failure_reason` metric.

GUS-W-8059892.


runesoerensen (approved):

bin/steps/nltk
```diff
@@ -21,17 +20,30 @@ EXPORT_PATH="${BUILDPACK_DIR}/export"
...
 
-		readarray -t nltk_packages <"$nltk_packages_definition"
+		readarray -t nltk_packages <"${nltk_packages_definition}"
 		output::step "Downloading NLTK packages: ${nltk_packages[*]}"
 
-		python -m nltk.downloader -d "$BUILD_DIR/.heroku/python/nltk_data" "${nltk_packages[@]}" | output::indent
-		set_env NLTK_DATA "/app/.heroku/python/nltk_data"
+		nltk_data_dir="/app/.heroku/python/nltk_data"
 
+		if ! python -m nltk.downloader -d "${nltk_data_dir}" "${nltk_packages[@]}" |& output::indent; then
```
colincasey, line 33: @edmorley the previous code downloaded to `$BUILD_DIR/.heroku/python/nltk_data` but the new code downloads to `/app/.heroku/python/nltk_data`. Won't that be a problem?

edmorley: No since `/app/.heroku/python` is a symlink to the equivalent build location (it just results in a nicer message in teh build log, as seen in the test fixture update) :-)

colincasey (approved):

Issue #1691, opened 2024-11-08
Fail the build if multiple package manager files detected


edmorley:
Currently the buildpack will use whichever package manager it finds first, if the files of multiple package managers are found.

This occasionally results in support tickets where the user believes the build to not be installing dependencies correctly, when in fact they are adding dependencies to the wrong package manager file.

To avoid this, we should make the presence of multiple package manager files an error.

We'll first want to add a warning so the change is not a surprise. (Particularly with the recent Poetry support addition, there will be apps using the third party Poetry buildpack that have both a `poetry.lock` and the generated `requirements.txt` until they remove the third-party buildpack.)

GUS-W-14336556.

PR #1692, opened 2024-11-08
Add a warning when multiple package managers are found

edmorley:
Currently the buildpack will use whichever package manager it finds first, if the files of multiple package managers are found.

This occasionally results in support tickets where the user believes the build to not be installing dependencies correctly, when in fact they are adding dependencies to the wrong package manager file.

As such, we want to make the presence of multiple package manager files an error, but we're first adding a warning so it's not a surprise. (Plus with the recent Poetry support addition, there will be apps using the third party Poetry buildpack that have both a `poetry.lock` and the generated `requirements.txt` until they remove the third-party buildpack.)

Towards #1691.
GUS-W-17167927.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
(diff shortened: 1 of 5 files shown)


colincasey (approved):
Learning from my mistake by making the detection of multiple package managers a warning before making it an error 😅

PR #1708, opened 2024-12-02
Fix Connection Error

ShivaniKumar1:
The error being returned is misleading. The error that should be returned should be something like
output::error <<-EOF Error: Python Connection timeout error EOF meta_set "failure_reason" "connection-timeout"
The error "python version not found" has nothing to do with the connection being timeout.

### DIFF
--- a/bin/steps/python
+++ b/bin/steps/python
@@ -17,12 +17,9 @@ PYTHON_URL="${S3_BASE_URL}/python-${python_full_version}-ubuntu-${UBUNTU_VERSION
 # TODO: Update this message to be more specific once Python 3.8 support is dropped.
 if ! curl --output /dev/null --silent --head --fail --retry 3 --retry-connrefused --connect-timeout 10 "${PYTHON_URL}"; then
 	output::error <<-EOF
-		Error: Python ${python_full_version} isn't available for this stack (${STACK}).
-
-		For a list of the supported Python versions, see:
-		https://devcenter.heroku.com/articles/python-support#supported-runtimes
+		Error: Python Connection timeout
 	EOF
-	meta_set "failure_reason" "python-version-not-found"
+	meta_set "failure_reason" "connection-timeout"
 	exit 1
 fi
 


bin/steps/python
```diff
@@ -17,12 +17,9 @@ PYTHON_URL="${S3_BASE_URL}/python-${python_full_version}-ubuntu-${UBUNTU_VERSION
 # TODO: Update this message to be more specific once Python 3.8 support is dropped.
 if ! curl --output /dev/null --silent --head --fail --retry 3 --retry-connrefused --connect-timeout 10 "${PYTHON_URL}"; then
 	output::error <<-EOF
-		Error: Python ${python_full_version} isn't available for this stack (${STACK}).
-
-		For a list of the supported Python versions, see:
-		https://devcenter.heroku.com/articles/python-support#supported-runtimes
+		Error: Python Connection timeout
```
edmorley, line 20: Hi! Thank you for the PR.

I agree this error message needs improving (fixing this was in my longer term todo list), since it currently doesn't distinguish between the 404 case and other failure modes. 

However, unfortunately this new error message will also not be correct most of the time either. For example:
- Some Python versions are not available on some stacks (eg Python 3.8 is not available on Heroku-22 and newer)
- The download could fail for other reasons that aren't a timeout (for example an HTTP 500 from S3 being down)

As such, this change would cause the Python 3.8 tests here to fail:
https://github.com/heroku/heroku-buildpack-python/blob/07005f2a3f56c78d9ae51398378e7ae629aa1f54/spec/hatchet/python_version_spec.rb#L115-L127

To fix all cases here, we'll need to inspect the HTTP status code from curl and vary the error message accordingly. Or we could wait until Python 3.8 support is dropped (which will be happening soon), since at that point all Python versions are available on all stacks.

edmorley (changes requested):

edmorley:
Hi! I've resolved this as part of the refactoring in #1749. The two new error message wordings can be seen here:
https://github.com/heroku/heroku-buildpack-python/pull/1749/files#diff-3ab441aec8280f446184fccdfde98f33111904de09816f6e3f13a8b4d1229189R86-R132

Thank you again for the PR and the bug report :-)

PR #1715, opened 2024-12-06
Use Python 3.13 as the default Python version for new apps

edmorley:
Previously if a new app did not specify a Python version (e.g. via a `.python-version` file), then the buildpack would default to the latest Python 3.12 release.

Now the buildpack will use the latest Python 3.13 release.

This does not affect:
* existing apps that have already had a build using the previous default version (since that version is remembered from one build to the next, via the "version pinning" / "sticky versions" feature).

Use-cases involving new apps that aren't compatible with Python 3.13 will now need to specify Python 3.12 explicitly:
https://devcenter.heroku.com/articles/python-support#specify-a-python-version

However most apps should be compatible, since Python 3.13 was released a couple of months ago (2024-10-07), does not contain significant breaking changes over Python 3.12, and has been happily used by the python-getting-started project since 2024-10-10.

See:
https://docs.python.org/3.13/whatsnew/3.13.html

Note: When using Python 3.13, the buildpack does not install setuptools, wheel or the sqlite headers/CLI, hence those diffs to the test fixtures.

GUS-W-14846972.


joshwlewis (approved):

edmorley:
Default version mentions updated at:
https://devcenter.heroku.com/articles/python-support#default-python-version-for-new-apps
https://devcenter.heroku.com/articles/python-runtimes

edmorley:
https://devcenter.heroku.com/changelog-items/3080

PR #1718, opened 2024-12-11
Improve the error message shown for legacy Celery

edmorley:
pip 24.1+ rejects packages that use invalid versions in their metadata (such as Celery older than v5.2.1).

We now annotate such failures with additional help text, mentioning the package needs to be upgraded, and link to the changelog entry which has additional advice:
https://devcenter.heroku.com/changelog-items/3073

This will help give clues to users who haven't seen the changelog entry (as seen in a couple of recent support tickets).

See also:
https://github.com/heroku/heroku-buildpack-python/pull/1685#issuecomment-2516690313

GUS-W-17415804.


spec/hatchet/pip_spec.rb
```diff
@@ -285,4 +285,33 @@
...
+          remote:  !     Newer versions of pip reject packages that use invalid versions
+          remote:  !     in their metadata (such as Celery older than v5.2.1).
+          remote:  !     
+          remote:  !     Try upgrading to a newer version of the affected package.
+          remote:  !     
+          remote:  !     For more help, see:
+          remote:  !     https://devcenter.heroku.com/changelog-items/3073
+          remote: 
+          remote: 
+          remote:  !     Error: Unable to install dependencies using pip.
```
edmorley, line 308: Note: This duplicate "Error:" heading is an existing papercut with the warnings feature, that's on my longer term todo list to fix, but out of scope here. (It also occurs in the GDAL testcase above.)

Malax (approved):

Issue #1710, opened 2024-12-03
`bundled_pip_wheel_list[0]: unbound variable` error if a broken Python install exists in app source


edmorley:
Spotted via Honeycomb metrics and then reproduced locally...

If an app's source contains a file at `.heroku/python/bin/python`, then the build currently fails with:

```
-----> Using Python 3.9.20 specified in .python-version
-----> Using cached install of Python 3.9.20
/tmp/buildpack/lib/utils.sh: line 33: bundled_pip_wheel_list[0]: unbound variable
```

Apps should not have any contents at `.heroku/python/` (it's an internal location used by the buildpack), but I'm guessing perhaps some users may have downloaded an app's slug to rescue the source from an app years ago (from a Python version that doesn't have `ensurepip`?), and then extracted that directly into their Git repo without removing the Python binaries?

GUS-W-17386432.

edmorley:
The error message here has been improved in #1720 / #1724.

Issue #1704, opened 2024-11-25
Error if the Python buildpack has been run multiple times during the same build


edmorley:
From looking at [buildpack usage data](https://data.heroku.com/dataclips/rxotsanetozvjufoqvdeyzcpsdcz) there are a small number of apps which are accidentally running the Python buildpack multiple times during the build.

For example:

```
-----> Using buildpacks:
       1. heroku/python
       2. heroku/python
```

Or:

```
-----> Using buildpacks:
       1. https://github.com/heroku/heroku-buildpack-python.git
       2. https://github.com/heroku/heroku-buildpack-apt.git
       3. heroku/python
```

Running the Python buildpack multiple times:
1. Slows down the build
2. Means the buildpack has to defensively handle files/directories existing in the build dir in locations that should be empty during steps like [cache restoration](https://github.com/heroku/heroku-buildpack-python/blob/ca99c39f9e3627fff727788206f7fcc4ff064b7f/lib/cache.sh#L141-L145))
3. Makes it more error prone for the Python buildpack to do things like fail the build if `PYTHONHOME` is set (which is something that should never be set via app config vars) - since it could have been set by an earlier run of the Python buildpack

As such, we should add a warning+metrics if Python-buildpack related files are found in the build directory, which we then eventually upgrade to an error message (that explains how to remove the duplicate buildpack).

edmorley:
An initial warning has been added in #1724.

PR #1724, opened 2024-12-13
Add build environment validation checks

edmorley:
- Adds a warning if the Python buildpack has been run multiple times in the same build.
- Adds a warning if an existing `.heroku/python/` directory is found in the app source.
- Improves the error message shown if the buildpack is used on an unsupported stack. Previously the build would fail with a curl download error depending on the availability of assets on S3. This scenario can only happen outside of Heroku, or in an old buildpack is used with a brand new stack (eg for Heroku-26).

The two warnings are the first step towards #1704 and #1710, and will be turned into errors in January 2025.

Towards #1704 and #1710.
GUS-W-17386432.
GUS-W-17309709.

### DIFF
--- a/lib/checks.sh
+++ b/lib/checks.sh
@@ -1,0 +1,116 @@
+#!/usr/bin/env bash
+
+function checks::ensure_supported_stack() {
+	local stack="${1}"
+
+	case "${stack}" in
+		heroku-20 | heroku-22 | heroku-24)
+			return 0
+			;;
+		cedar* | heroku-16 | heroku-18)
+			# This error will only ever be seen on non-Heroku environments, since the
+			# Heroku build system rejects builds using EOL stacks.
+			output::error <<-EOF
+				Error: The '${stack}' stack is no longer supported.
+
+				This buildpack no longer supports the '${stack}' stack since it has
+				reached its end-of-life:
+				https://devcenter.heroku.com/articles/stack#stack-support-details-for-apps-using-classic-buildpacks
+
+				Upgrade to a newer stack to continue using this buildpack.
+			EOF
+			meta_set "failure_reason" "stack::eol"
+			exit 1
+			;;
+		*)
+			output::error <<-EOF
+				Error: The '${stack}' stack is not recognised.
+
+				This buildpack does not recognise or support the '${stack}' stack.
+
+				If '${stack}' is a valid stack, make sure that you are using the latest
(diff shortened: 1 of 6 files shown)


lib/checks.sh
```diff
@@ -0,0 +1,116 @@
...
+	# since app provided config vars haven't been exported to the environment here.)
+	if [[ -f "${build_dir}/.heroku/python/bin/python" && -v PYTHONHOME ]]; then
+		output::warning <<-EOF
+			Warning: The Python buildpack has already been run this build.
+
+			An existing Python installation was found in the build directory
+			from a buildpack run earlier in the build.
+
+			This normally means there are duplicate Python buildpacks set
+			on your on your app, which is not supported since it can cause
```
schneems, line 58:
```suggestion
			on your app, which is not supported since it can cause
```

lib/checks.sh
```diff
@@ -0,0 +1,116 @@
...
+			Check the buildpacks set on your app and remove any duplicate
+			Python buildpack entries:
+			https://devcenter.heroku.com/articles/managing-buildpacks#view-your-buildpacks
+			https://devcenter.heroku.com/articles/managing-buildpacks#remove-classic-buildpacks
+
+			If you have a use-case that requires duplicate buildpacks,
+			please comment on:
+			https://github.com/heroku/heroku-buildpack-python/issues/1704
+
+			In January 2025 this warning will be made an error.
```
schneems, line 70: Not blocking comment: This is a pretty aggressive time scale, especially for devs out on holidays and not actively deploying. I feel like we are validated in encouraging people to not do this though, so I'm for an accelerated timescale.

When we change the error to a warning, I would like an env var as an "escape valve" to accomadate not blocking people who (may) have a valid reason we've not yet thought about but haven't had a chance to comment on the issue. For example: An env var to turn the error into a warning or directing people to use an older version of the biuldpack (though that risks people being accidentally encouraged to get stuck on an old buildpack version if they forget about it). A thought.

edmorley: I'll be basing the eventual decision on the additional metics.

However, I already know this is causing real errors (see #1704), and also know very few apps run duplicate buildpacks ([dataclip](https://data.heroku.com/dataclips/rxotsanetozvjufoqvdeyzcpsdcz)).

schneems (approved):
Looks good. Reminds me of https://github.com/heroku/heroku-buildpack-ruby/pull/552.

PR #1732, opened 2025-01-07
Remove Python 3.8 support + deprecate Python 3.9

edmorley:
Python 3.8 reached upstream EOL on 7th October 2024:
https://devguide.python.org/versions/#supported-versions

The Python version support policy is that supported versions follows the upstream EOL lifecycle:
https://devcenter.heroku.com/articles/python-support#python-version-support-policy

As such, the buildpack has been showing a deprecation warning with a scheduled removal date since December 2023:
- https://github.com/heroku/heroku-buildpack-python/pull/1515
- https://github.com/heroku/heroku-buildpack-python/pull/1721
- https://devcenter.heroku.com/changelog-items/2768

(Plus Python 3.8 binaries have never been available for Heroku-22 or Heroku-24, meaning it was only ever supported on Heroku-20 and older.)

Dropping support for Python 3.8 also unblocks upgrading to Poetry v2 (which has already dropped 3.8 support).

Apps using Python 3.8 that aren't able to upgrade immediately will need to pin to an older buildpack version temporarily (which will work until Heroku-20 EOLs).

A deprecation warning has now also been added for Python 3.9.

GUS-W-17472311.
GUS-W-14846945.


Malax (approved):

edmorley:
https://devcenter.heroku.com/changelog-items/3094
https://devcenter.heroku.com/changelog-items/3095
https://devcenter.heroku.com/articles/python-support#supported-python-versions

PR #1787, opened 2025-05-05
Warn if an app doesn't specify a Python version

edmorley:
Adds a warning when an app doesn't specify a Python version and instead relies upon the buildpack's default/cached version.

Relying on the default Python version isn't recommended, since:
1. The default version may not match the version used locally, in CI, or in other instances of the app (including eg review apps).
2. The default version changes over time, and while the classic buildpack does reuse the major version cached from the last build, this is lost is the cache is manually cleared by the user, or when migrating from one app to another (including when migrating to a new Fir app).

In addition, using a .python-version file will be mandatory for apps that use the upcoming uv support (since uv has native .python-version file support, and without that file, the UX around conflicting versions would be worse) - so it makes sense that we should at least warn when not using one for other package managers too.

GUS-W-18439666.


lib/python_version.sh
```diff
@@ -432,7 +432,23 @@ function python_version::warn_if_python_version_file_missing() {
...
+				Your app doesn't specify a Python version and so the buildpack
+				picked a default version for you.
+
+				Relying on this default version isn't recommended, since it
+				can change over time and may not be consistent with your local
+				development environment, CI or other instances of your app.
+
+				Please configure an explicit Python version for your app.
+
+				${instructions}
```
schneems, line 447: Only recommendation would be adding some git instructions like https://github.com/heroku/heroku-buildpack-ruby/blob/f506e9ae3931bbce010fbcaa37c59fd26638f3c0/lib/language_pack/helpers/bundler_wrapper.rb#L225-L226

Otherwise, they might have the file locally, but forget to check it in.

schneems (approved):

lib/python_version.sh
edmorley: Ah great idea - thank you!

edmorley: @schneems Ah great idea - thank you!

edmorley: Added in 2211bb8d4bb6404aa795ba346579eefa288c606e

Issue #1700, opened 2024-11-18
allow poetry config settings


claireramming:
I have several applications that use poetry and several of the packages in each app need to be installed via a private repository. Previously I was able to install these in my apps using a requirements.txt that installed poetry, then a post_compile script that set poetry config vars for http basic sign-in to our private repository before calling poetry install. 

With the new buildpack updates that allow poetry to be used out of the box, I'm struggling to figure out how to correctly pass my private repository user/password since I can no longer call my config command directly before `poetry install` is called (I don't think pre_compile will work here since poetry needs to be installed first). According to poetry I could use a config.toml file, but since one of the vars is a pw I don't want to expose that. Another option via poetry is that I should be able to set ENV vars to replace the config command that I used to call before running poetry install ([poetry doc here](https://python-poetry.org/docs/configuration/#http-basicnameusernamepassword)) but poetry in Heroku does not seem to be picking up on my POETRY_ env vars, even though from what I can tell, it should be passing them... Previously with the post_compile config command I was setting the user/pass also with env vars (ex: `poetry config http-basic.my-pypi "$PYPI_USERNAME" "$PYPI_PASSWORD"`) and it was working fine.

edmorley:
@claireramming Hi! Thank you for filing this :-)

The Python buildpack hasn't historically exposed app config vars (apart from `PIP_EXTRA_INDEX_URL`) when running the pip/Pipenv package manager install steps, since some apps have broken env vars set (for example invalid values for `PYTHONHOME`, `PYTHONPATH` etc) that would cause the package manager to fail in hard to debug ways. As such, when adding support for Poetry I chose to match that approach for parity.

However, I would like to see if we can change this in the future (which would allow you to use the `POETRY_HTTP_BASIC_<NAME>_USERNAME` and `POETRY_HTTP_BASIC_<NAME>_PASSWORD` env vars) - it will just require careful thought about what env vars are potentially problematic - and for the buildpack to either filter those out, or hard error with an error message saying those broken env vars should be unset. 

(The buildpack already has some rudimentary filtering [here](https://github.com/heroku/heroku-buildpack-python/blob/ca99c39f9e3627fff727788206f7fcc4ff064b7f/vendor/buildpack-stdlib_v8.sh#L41), however, we also need to check for additional env vars, plus cover the case where env vars have been set by a [duplicate Python buildpack](https://github.com/heroku/heroku-buildpack-python/issues/1704).)

[...]

Issue #1451, opened 2023-05-26
Trouble using config vars in requirements.txt


ericwb:
Unsure if I'm somehow doing something wrong or this is by design. According to Pip's documentation, I can include a private repo in my requirements.txt as a dependency. Since the repo is private and hosted on GitHub, I need to provide a personal access token. Of course, I do not want that token included in source code, so I configure it in Heroku's Config Vars. But it doesn't seem to work for me. However works fine when running locally.

For example, requirements.txt:
```
aiohttp
cryptography
cachetools
sqlalchemy==1.4.46
git+https://${GITHUB_TOKEN}@github.com/user/repo.git@main
```

This is the error I get from Heroku. It appears, it does not resolve the GITHUB_TOKEN environmental variable that is defined in the Config Vars.
```
-----> Installing requirements with pip
       Collecting git+https://****@github.com/user/repo.git@main (from -r requirements.txt (line 9))
         Cloning https://****@github.com/user/repo.git (to revision main) to /tmp/pip-req-build-imw3lwma
         Running command git clone --filter=blob:none --quiet 'https://****@github.com/user/repo.git' /tmp/pip-req-build-imw3lwma
         fatal: could not read Password for 'https://${GITHUB_TOKEN}@github.com': No such device or address
         error: subprocess-exited-with-error
```

Is there a workaround or something I'm missing? Thanks

https://pip.pypa.io/en/stable/reference/requirements-file-format/#using-environment-variables

ericwb:
I should also add that the Heroku docs state:
"Anything that works with a standard pip requirements file will work as expected on Heroku."

However, following that statement, it gives an example that is unsupported on GitHub. GitHub no longer supports user/password based authentication.

```
git+https://user:password@github.com/nsa/secret.git
```

https://devcenter.heroku.com/articles/python-pip#git-backed-distributions

ericwb:
I guess this issue has come up several times in different forms:
https://github.com/heroku/heroku-buildpack-python/pull/417
https://github.com/heroku/heroku-buildpack-python/pull/467
https://github.com/heroku/heroku-buildpack-python/pull/639

The only workaround I seem to have is hosting another python package index and using PIP_EXTRA_INDEX_URL. Which is a hacky option for one repo.

ericwb:
This link seems to provide a good option on how to modify heroku-buildpack-python to utilize and acceptlist and denylist for env vars. Maybe this would be a worthy consideration as a fix?

[...]

PR #1791, opened 2025-05-07
Add support for the package manager uv

### LINKED ISSUE #1616: Support the uv package manager
`uv` is a drop-in replacement for `pip` which runs very quickly. We've replaced `pip` with `uv` locally and have noticed a significant speed up in our Python build time in our development and CI environments. By adopting support for `uv`, we anticipate our Heroku builds would complete ~5 minutes more quickly. 

Because `uv` is a drop-in replacement, we found adoption to be very straightforward in our case. `pip install -r requirements.txt` becomes `uv pip install -r requirements.txt`. The promise of significantly improved performance with a very simple migration path was enough to convince us to give it a try internally.

Happy to attempt a PR if that would be welcome/well received.

edmorley:
Adds support for the package manager uv:
https://docs.astral.sh/uv/

...which has (very justifiably!) become extremely popular in the last year:

uv-star-history-2025-05-07

This specifically adds support for uv's project mode (`uv sync`), which uses a lockfile rather then a `requirements.txt` file:
https://docs.astral.sh/uv/guides/projects/

(While uv does have a pip compatible mode that supports requirements files, using a lockfile is a best practice for deploying an app, and not using one means losing most of the benefits of using uv.)

Apps must have a `pyproject.toml`, `uv.lock` and `.python-version` file (all of which are created by default by `uv init` followed by either `uv lock` or any of the other commands that sync/update the lockfile).

We don't support `runtime.txt` (or omitting the `.python-version` file) when using uv, since otherwise it leads to a much worse UX in several cases. (For example, when the `runtime.txt` version or buildpack default version differs from `requires-python` in the `pyproject.toml` file, particularly when `requires-python` happens to also match the distro Python in the underlying base image.)

For now, if a `requirements.txt`, `Pipfile` or `poetry.lock` are found, they will take precedence over `uv.lock` for backwards compatibility (currently a warning is shown, but in the future this will become an error). This means users of the third-party [heroku-buildpack-uv](https://github.com/dropseed/heroku-buildpack-uv) will need to remove that buildpack in order to use the new native uv support, since it exports a `requirements.txt` file during the build which will take precedence.

For standard (non-CI) Heroku builds, the buildpack installs dependencies using `uv sync --locked --no-default-groups`, to ensure development/test dependencies are not installed.

We don't use `--no-dev` instead, since that only excludes the `dev` group, and some apps may be using other group names and have added them to uv's `default-groups` in `pyproject.toml` so that local development workflows work with a plain `uv sync` with no additional args. ([PEP 735](https://peps.python.org/pep-0735/) doesn't actually prescribe recommended group names, and in fact uses `test` in its example rather than `dev`, so non-standard names may be common.)

For Heroku CI builds, the buildpack installs dependencies using `uv sync --locked`, which will install all default groups.

For more on uv's dependency group behaviour, see:
https://docs.astral.sh/uv/concepts/projects/dependencies/#dependency-groups

For now, we don't expose user provided env vars to `uv sync` (similar to the behaviour for other package managers), however, will start doing so as part of addressing #1451 and #1700.

uv is installed into the build cache rather than the slug, so is not available at run-time (since it's not typically needed at run-time and doing so reduces the slug size). Both Python and the entrypoints of installed dependencies are available on `PATH`, so use of `uv run` is not required at run-time to use dependencies in the environment.

Closes #1616.
GUS-W-17310796.

---

Note: This repository is for the classic Python buildpack used by [Cedar-generation](https://devcenter.heroku.com/articles/generations#cedar) apps on Heroku. For uv support for [Fir-Generation](https://devcenter.heroku.com/articles/generations#fir) apps, see the Python CNB's [tracking issue](https://github.com/heroku/buildpacks-python/issues/248) instead.


edmorley:
Example of using this with the Python Getting Started guide repo (with a demo branch that has had it's `requirements.txt` replaced with a `pyproject.toml` and `uv.lock`):

```
$ git clone --branch edmorley/uv-demo https://github.com/heroku/python-getting-started && cd python-getting-started
$ heroku create --buildpack 'https://github.com/heroku/heroku-buildpack-python#uv-support'
$ git push heroku HEAD:main
...
remote: -----> Building on the Heroku-24 stack
remote: -----> Using buildpack: https://github.com/heroku/heroku-buildpack-python#uv-support
remote: -----> Python app detected
remote: -----> Using Python 3.13 specified in .python-version
remote: -----> Installing Python 3.13.3
remote: -----> Installing uv 0.7.2
remote: -----> Installing dependencies using 'uv sync --locked --no-default-groups'
remote:        Resolved 11 packages in 0.53ms
remote:        Downloading django (7.9MiB)
remote:        Downloading brotli (2.8MiB)
remote:         Downloading brotli
remote:         Downloading django
remote:        Prepared 9 packages in 499ms
remote:        Installed 9 packages in 360ms
remote:        Bytecode compiled 979 files in 695ms
remote:         + asgiref==3.8.1
remote:         + brotli==1.1.0
remote:         + dj-database-url==2.3.0
remote:         + django==5.2.1
remote:         + gunicorn==23.0.0
remote:         + packaging==25.0
remote:         + sqlparse==0.5.3
remote:         + typing-extensions==4.13.2
remote:         + whitenoise==6.9.0
remote: -----> $ python manage.py collectstatic --noinput
remote:        WARNING:root:No DATABASE_URL environment variable set, and so no databases setup
remote:        1 static file copied to '/tmp/build_9566b80e/staticfiles', 1 post-processed.
remote:
remote: -----> Discovering process types
remote:        Procfile declares types -> web
remote:
remote: -----> Compressing...
remote:        Done: 24.1M
remote: -----> Launching...
remote:        Released v3
remote:        https://mysterious-fjord-13174-90ef5e4473c3.herokuapp.com/ deployed to Heroku

$ curl -sSf https://mysterious-fjord-13174-90ef5e4473c3.herokuapp.com/ | head -n 4
<!DOCTYPE html>
<html>
<head>
  <title>Python Getting Started on Heroku</title>
```

lib/uv.sh
```diff
@@ -0,0 +1,165 @@
...
+		export UV_PROJECT_ENVIRONMENT="${python_home}"
+		export UV_PYTHON_DOWNLOADS="never"
+		export UV_PYTHON_PREFERENCE="only-system"
+	EOF
+
+	# As a performance optimisation, uv attempts to use hardlinks instead of copying files from its
+	# download cache into site-packages, and will emit a warning if it has to fall back to copying.
+	# By default uv stores its cache under `$HOME/.cache`, and for standard Heroku builds `$HOME` is
+	# `/app`, which is on a different filesystem mount to the build directory (which is under `/tmp`),
+	# meaning hardlinks can't be used. To avoid this we tell uv's to store its cache in `/tmp`.
```
colincasey, line 95:
```suggestion
	# meaning hardlinks can't be used. To avoid this we tell uv to store its cache in `/tmp`.
```

spec/hatchet/uv_spec.rb
```diff
@@ -0,0 +1,361 @@
...
+        REGEX
+      end
+    end
+  end
+
+  # This tests the error message when there is no .python-version file, and in particular the case where
+  # the buildpack's default Python version is not compatible with `requires-python` in pyproject.toml.
+  # (Since we must prevent uv from downloading its own Python or using system Python, and also
+  # want a clearer error message than using `--python` or `UV_PYTHON` would give us).
+  context 'when there is no .python-version file' do
```
colincasey, line 244: The match for this case is `default | cached)`, right? This seems to cover the `default` case. How does the `cached` case get triggered?

edmorley: `cached` is where the Python version was inferred from the previous build:
https://github.com/heroku/heroku-buildpack-python/blob/819fb61774b66abddbd998db5c44f12d0d03afb1/lib/python_version.sh#L82

xref:
https://github.com/heroku/heroku-buildpack-python/blob/819fb61774b66abddbd998db5c44f12d0d03afb1/lib/python_version.sh#L26-L43

colincasey: Then does this warrant an additional test with a rebuild to trigger this case or are you satisfied with the current coverage?

edmorley: The `default` vs `cached` cases are covered pretty extensively already in the non-uv tests, and so this branch of the case statement is already tested for non-uv, and the uv-specific error is tested via the `default` case. You're correct there isn't an explicit test for this permutation (ie 3 out of the 4 permutations for this error/warning are tested). The cached tests are quite a bit slower (since they need to do a repeat build), so I've generally tried to avoid too much overlap between the test cases - but I can add another explicit test case for it now.

lib/uv.sh
edmorley: Thank you - typo fixed in f8470f6b536172216f567f1cfe43f15980ec58e9

spec/hatchet/uv_spec.rb
edmorley: Additional test case added as part of f8470f6b536172216f567f1cfe43f15980ec58e9

edmorley:
Dev Center updates:
- https://devcenter.heroku.com/articles/python-behavior-in-heroku ([diff](https://devcenter.heroku.com/admin/articles/diff/8822/v5..v8))
- https://devcenter.heroku.com/articles/heroku-ci#python ([diff](https://devcenter.heroku.com/admin/articles/diff/3562/v190..v191))

Changelog announcement:
https://devcenter.heroku.com/changelog-items/3238

PR #1840, opened 2025-07-23
Update Pipenv to 2025.0.4 and improve installation/caching

edmorley:
* Updates Pipenv from 2024.0.1 to 2025.0.4 ([changelog](https://github.com/pypa/pipenv/blob/main/CHANGELOG.md#202504-2025-07-07)).
* Switches to installing Pipenv into its own virtual environment, so that it and its dependencies don't leak into the app environment.
* Stops installing pip (in addition to Pipenv) when the chosen package manager is Pipenv, to match the behaviour when using Poetry or uv.
* Forces the build cache to be cleared if the contents of `Pipfile.lock` has changed since the last build. This is required to work around `pipenv {install,sync}` [by design](https://github.com/pypa/pipenv/issues/3365) not uninstalling packages if they are removed from the lockfile (?!). We also can't use `pipenv clean` since it doesn't work with `--system` / `PIPENV_SYSTEM`.

Apps that have been inadvertently depending on Pipenv's own dependencies implicitly, will need to add those dependencies explicitly to their `Pipfile` and regenerate `Pipfile.lock`. 

In particular, if you see the error `ModuleNotFoundError: No module named 'pkg_resources'` it means the app (or one of its dependencies) relies upon the `setuptools` package, but that package is missing from `Pipfile` / `Pipfile.lock` and must now be added there.

Similarly, any apps that for some reason use pip later in the build in addition to Pipenv, will need to either add pip as an explicit dependency in their `Pipfile`, or else for adhoc use-cases can temporarily install pip using `python -m ensurepip --default-pip`.

Supersedes #1828.
GUS-W-17482289.
GUS-W-17482412.
GUS-W-19116903.


edmorley:
https://devcenter.heroku.com/changelog-items/3324

edmorley:
This exposed an underlying bug in Pipenv's `--system` implementation, for which I have a workaround in #1842.

PR #1875, opened 2025-08-19
Remove support for `BUILDPACK_S3_BASE_URL`

edmorley:
Since:
- Metrics show it to be unused on Heroku.
- It's not supported by the Python CNB, so removing it improves parity.
- As a feature it's pretty fragile, since unless someone managed to keep up with all changes we make to our S3 bucket, builds would break every time we bump versions. (And as such, anyone using it should be carefully curating the buildpack versions, and so can patch the buildpack as part of that repackaging/automation process.)

GUS-W-19372843.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Removed support for `BUILDPACK_S3_BASE_URL`. ([#1875](https://github.com/heroku/heroku-buildpack-python/pull/1875))
 
 ## [v301] - 2025-08-18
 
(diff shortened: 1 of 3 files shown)


dzuelke (approved):

schneems (approved):
Pretty sure the main use case is off heroku. Like running via dokku. This is the PR that added support to Ruby https://github.com/heroku/heroku-buildpack-ruby/pull/238. The way it's implemented in Ruby, a heroku customer CANNOT set it via `heroku config`, you would have to set it via another buildpack or externally (like the process running the buildpack process).

Commenting for extra context.

edmorley:
@schneems Yeah I realise it's primarily for non-Heroku. It was also not settable via `heroku config` (this is explained in the code comment removed in this PR). However, had someone on Heroku been using it, they could have set it via an earlier buildpack, hence double checking via metrics.

PR #1878, opened 2025-08-20
Refactor buildpack data store and `bin/report`

edmorley:
The Node.js buildpack has a set of Bash helper utils for managing a key-value store of build data/metadata, that could be added to during the build, and then output when `bin/report` was run. When the Python buildpack added support for `bin/report`, its implementation was based on the Node.js one, with some further small improvements.

However, there remains a number of issues with the design:
1. It handles different data types (string vs int vs float vs bool) at `bin/report` time, so requires a hardcoded mapping of values which can (and does) get out of sync. (Plus doesn't work well for some cases, like jvm-common being wrapped by the other JVM related buildpacks.)
2. It tries to escape newlines and carriage returns, but there are a number of edge cases still not handled properly which result in some fields not being sent to Honeycomb properly.

In order to work around the above, Manuel created a jq based approach for use by the JVM buildpacks:
https://github.com/heroku/heroku-buildpack-jvm-common/blob/4fd6e4f8f258a5766a8b2e9ffec2690675b00879/lib/metrics.sh

However, that doesn't have the full API of the existing solution (it's missing the ability to fetch previous values; something which Python needs), and I spotted a few more places where things can be optimised.

In addition, now that we're using jq and JSON internally in the buildpack, it makes sense for us to update cytokine to support the build report being emitted as JSON too (in addition to YAML), particularly since cytokine then submits the report to Vacuole as JSON afterwards.

And in fact, since JSON is roughly a subset of YAML (at least for the simple data types and schema we're using for the build report), we can start emitting the build report as JSON now, rather than needing to wait for the cytokine change to land first. (We'll still want to update cytokine to use a JSON parser though, just to avoid any edge cases in the future).

The new implementation:
- Sets the data type at the call site where the build data is set, rather than relying on a hardcoded mapping at time of retrieval.
- Uses jq to do the string escaping, so much less fragile than the hand-rolled implementation.
- Has been rewritten to avoid unnecessary processes calls, redirections, and patterns that required disabling a number of shellcheck warnings.

I've confirmed that metrics still appear in Honeycomb with the changes here.

After this merges, I'll open PRs against some of the other language buildpacks that don't yet implement `bin/report`, using the improved implementation from here (with simplifications since other buildpacks won't need the legacy data store handling that Python needs to support caches written by older buildpack versions). And once Manuel is back, I'll discuss with him backporting some of the improvements to the JVM buildpacks too.

GUS-W-19344754.

### DIFF
--- a/bin/compile
+++ b/bin/compile
@@ -20,10 +20,10 @@ BUILDPACK_DIR=$(cd "$(dirname "$(dirname "${BASH_SOURCE[0]}")")" && pwd)
 
 source "${BUILDPACK_DIR}/bin/utils"
 source "${BUILDPACK_DIR}/lib/utils.sh"
+source "${BUILDPACK_DIR}/lib/build_data.sh"
 source "${BUILDPACK_DIR}/lib/cache.sh"
 source "${BUILDPACK_DIR}/lib/checks.sh"
 source "${BUILDPACK_DIR}/lib/hooks.sh"
-source "${BUILDPACK_DIR}/lib/build_data.sh"
 source "${BUILDPACK_DIR}/lib/output.sh"
 source "${BUILDPACK_DIR}/lib/package_manager.sh"
 source "${BUILDPACK_DIR}/lib/pip.sh"


lib/metadata.sh
```diff
@@ -4,75 +4,140 @@
...
+# ```
+function metadata::set_raw() {
+	local key="${1}"
+	local value="${2}"
+	metadata::_set "${key}" "${value}" "false"
 }
 
-# Similar to mtime from buildpack-stdlib
-meta_time() {
+function metadata::_set() {
```
schneems, line 111: (optional) Suggest adding docs and an example for completeness. Even if it's an internal function it helps knowing the intent and seeing an example of the API without having to read the body.

```suggestion
# Internal helper to write a key/value pair to the metadata store.
# Takes a key, value, and a boolean flag indicating whether the value needs to be quoted.
#
# Usage:
# ```
# metadata::_set "foo_string" "quote me" "true"
# metadata::_set "bar_number" "99" "false"
# ```
function metadata::_set() {
```

lib/metadata.sh
```diff
@@ -4,75 +4,140 @@
...
 
-# Retrieve a value from a previous build if it exists
-# This is useful to give the user context about what changed if the
-# build has failed. Ex:
-#   - changed stacks
-#   - deployed with a new major version of Node
-#   - etc
-meta_prev_get() {
-	kv_get "${PREVIOUS_BUILD_DATA_FILE}" "${1}"
+# Returns the current time in milliseconds since the UNIX Epoch.
```
schneems, line 136:
```suggestion
# Returns the current time in milliseconds since the UNIX Epoch.
#
# Usage:
# ```
# local dependencies_install_start_time=$(metadata::current_unix_time_ms)
# # ... some operation ...
# metadata::set_duration "dependencies_install_duration" "${dependencies_install_start_time}"
# ```
```

lib/metadata.sh
```diff
@@ -4,75 +4,140 @@
...
+	date +%s%3N
 }
 
-log_meta_data() {
-	# print all values on one line in logfmt format
-	# https://brandur.org/logfmt
-	# the echo call ensures that all values are printed on a single line
-	# shellcheck disable=SC2005,SC2046,SC2312
-	echo $(kv_list "${BUILD_DATA_FILE}")
+function metadata::print_bin_report_json() {
```
schneems, line 141:
```suggestion
# Prints the contents of the metadata store in sorted JSON format.
#
# Usage:
# ```
# metadata::print_bin_report_json
# ```
function metadata::print_bin_report_json() {
```

lib/metadata.sh
```diff
@@ -4,75 +4,140 @@
...
-	kv_clear "${BUILD_DATA_FILE}"
+	# Older versions of this buildpack used a `key=value` format file instead of JSON,
+	# so we need to support this file format/location too, so older caches can be read.
+	# We check for this file first, so that we correctly handle the case where an app
+	# downgraded and then re-upgraded buildpack version, so has both files in the cache.
+	if [[ -f "${LEGACY_BUILD_DATA_FILE}" ]]; then
+		# The legacy file contains one entry per line, of form `key=value`. Entries were written in an
+		# append-only manner so there could be duplicate entries for each key, so we return only the
+		# last matching entry in the file. The empty string is returned if the key wasn't found.
+		tac "${LEGACY_BUILD_DATA_FILE}" | { grep --perl-regexp --only-matching --max-count=1 "^${key}=\K.*$" || true; }
```
schneems, line 57: (optional) Suggestion: Exercise this `get_previous` legacy logic under test. 

I poked around, and it looks like there's no unit test capability right now. You could add something like https://github.com/heroku/heroku-buildpack-ruby/blob/a6d124492aafe7f7ee0dd42d7a79284b6552d675/spec/unit/bash_functions_spec.rb#L94 that executes the code locally (but requires OS command parity). Or you could have an inline buildpack that writes some old data to that file, load the `metadata.sh`, and then `get_previous` and echo them (doesn't need to run the Python buildpack, just `metadata.sh` contents).

schneems (approved):

lib/metadata.sh
edmorley: (values like the previous uv version come from the metadata store, and are listed in the build output in the test assertion, proving it read the old value correctly: https://github.com/heroku/heroku-buildpack-python/blob/3c27de9acc7a7bbcaf27d281e1381cd81c793999/spec/hatchet/uv_spec.rb#L106)

PR #1881, opened 2025-08-22
Use EPOCHREALTIME in build_data timing logic

dzuelke:
Built-in Bash (5.0+) variable, a float with microseconds precision. Faster, simpler, and more portable (no longer requires GNU date).

Logging it with six digits worth of precision instead of the previous three, but a float is a float ;)

(without a precision indicator, awk gives the usual hand-wavy float result of possibly a few more digits, we just discard that)

GUS-W-19404492


lib/build_data.sh
```diff
@@ -43,20 +43,20 @@ function build_data::set_string() {
...
 # build_data::set_duration "dependencies_install_duration" "${dependencies_install_start_time}"
 # ```
 function build_data::set_duration() {
 	local key="${1}"
 	local start_time="${2}"
 	local end_time duration
-	end_time="$(build_data::current_unix_time_ms)"
-	duration="$(awk -v start="${start_time}" -v end="${end_time}" 'BEGIN { printf "%.3f", (end - start)/1000 }')"
+	end_time="$(build_data::current_unix_realtime)"
+	duration="$(awk -v start="${start_time}" -v end="${end_time}" 'BEGIN { printf "%f", (end - start) }')"
```
edmorley, line 59:
```suggestion
	duration="$(awk -v start="${start_time}" -v end="${end_time}" 'BEGIN { print (end - start) }')"
```

lib/build_data.sh
```diff
@@ -134,17 +134,19 @@ function build_data::get_previous() {
...
 # build_data::set_duration "dependencies_install_duration" "${dependencies_install_start_time}"
 # ```
-function build_data::current_unix_time_ms() {
-	date +%s%3N
-}
+function build_data::current_unix_realtime() (
+	# we use a subshell with LC_ALL=C to ensure output format is not affected by system locale
+	LC_ALL=C
+	echo "${EPOCHREALTIME}"
+)
```
edmorley, line 149: (Just since the subshell was too hard to spot before, since the brackets looked like braces, it took me several times of looking at it to see them)

```suggestion
function build_data::current_unix_realtime() {
	# We use a subshell with `LC_ALL=C` to ensure the output format isn't affected by system locale.
	(
		LC_ALL=C
		echo "${EPOCHREALTIME}"
	)
}
```

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - Updated uv from 0.8.9 to 0.8.13. ([#1880](https://github.com/heroku/heroku-buildpack-python/pull/1880))
+- Use `EPOCHREALTIME` (Bash 5.0+) with microsecond precision for buildpack data store timing logic ([#1881](https://github.com/heroku/heroku-buildpack-python/pull/1881))
```
edmorley, line 6: (I typically try and use past tense for the changelog entries)

```suggestion
- Switched to Bash 5.0's `EPOCHREALTIME` for buildpack data store timing logic. ([#1881](https://github.com/heroku/heroku-buildpack-python/pull/1881))
```

edmorley (approved):
Thank you! Good to switch to something more lightweight than the `date` subprocess call :-)

lib/build_data.sh
edmorley: The `print` causes less precision, eg:

```
root@d4d36b9d6a56:/# awk -v start="${start_time}" -v end="${end_time}" 'BEGIN { print (end - start) }'
13.2063
```

So it would be better to either switch back to `printf`:

```
root@d4d36b9d6a56:/# awk -v start="${start_time}" -v end="${end_time}" 'BEGIN { printf "%f", (end - start) }'
13.206290
```

Or use `OFMT` with `print`:
```
root@d4d36b9d6a56:/# awk -v start="${start_time}" -v end="${end_time}" -v OFMT="%f" 'BEGIN { print (end - start) }'
13.206290
```

Given that we don't want the trailing newline (albeit the command substitution removes it for us), lets go with `printf` again (plus it saves people having to understand `OFMT`).

PR #1924, opened 2025-10-06
Switch from GPG to Sigstore for Python source verification

edmorley:
The build scripts that download and compile the Python source archives (for upload to S3, where they are then consumed by the buildpack during customer builds) currently use GPG to verify the Python source archive downloads.

However, use of PGP signatures for Python artifact verification was deprecated previously in PEP 761, in favour of Sigstore:
https://peps.python.org/pep-0761/
https://www.python.org/downloads/metadata/sigstore/

Until now the PGP signatures have still been available for all stable releases, however, as of Python 3.14 (due to be released this week), Sigstore will be the only supported verification mechanism:
https://docs.python.org/3.14/whatsnew/3.14.html#whatsnew314-no-more-pgp

As such, we must now switch over to Sigstore.

We use the `cosign` CLI for verification since it's a standalone binary available via a Docker image, rather than the Python `sigstore` CLI which requires a Python environment (and so would need pip, venv etc, and more setup to ensure it stays isolated from the Python we're trying to build).

See:
- https://www.python.org/downloads/metadata/sigstore/
- https://docs.sigstore.dev/cosign/system_config/installation/#container-images
- https://docs.sigstore.dev/cosign/verifying/verify/
- https://github.com/sigstore/cosign/blob/main/doc/cosign_verify-blob.md

GUS-W-18244071.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,6 +13,14 @@ updates:
         update-types:
           - "minor"
           - "patch"
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    labels:
+      - "dependencies"
+      - "docker"
+      - "skip changelog"
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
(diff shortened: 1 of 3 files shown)


edmorley:
Dry-run (non-upload) builds of the latest patch release of each major - to confirm the release manager identities are correct for each major:
- 3.9: https://github.com/heroku/heroku-buildpack-python/actions/runs/18278574141
- 3.10: https://github.com/heroku/heroku-buildpack-python/actions/runs/18278569193
- 3.11: https://github.com/heroku/heroku-buildpack-python/actions/runs/18278561315
- 3.12: https://github.com/heroku/heroku-buildpack-python/actions/runs/18278556177
- 3.13: https://github.com/heroku/heroku-buildpack-python/actions/runs/18278536028

Malax (approved):

PR #1925, opened 2025-10-06
Stop compiling Python using `--with-system-expat`

edmorley:
The upstream Python test suite (which gets run when compiling with PGO enabled) fails with the `libexpat` in Ubuntu 22.04.

In #1661, I previously added what I hoped would be a temporarily workaround until the failures were fixed upstream, however, the Python maintainers say they don't guarantee compatibility with distro `expat`, and that it's up to us to test for compatibility and patch if we want to use the distro version. However, this isn't viable given that we're neither a Linux distro maintainer, a CPython maintainer or an expat maintainer.

Instead, like the upstream Docker Hub Python images (who were also affected by this issue), we will switch the `expat` bundled within the CPython sources, which is actually what the upstream CPython project tests in its CI. This means users won't get security updates for free via the base image, and will instead need to update their Python patch versions instead as newer versions are vendored in CPython. However, this is the least worst alternative for now.

I'm doing this now, since otherwise I'll need to generate another patch series for the soon to be released Python 3.14.

Note: This change only affects Python versions compiled/released after this merges. Existing Python versions on S3 are unaffected for now (unless they ever get recompiled in the future).

For more details, see:
https://github.com/python/cpython/issues/125067#issuecomment-2517445480

GUS-W-17414073.


edmorley:
Dry-run (non-upload) builds of the latest patch release of each major - to confirm the release manager identities are correct for each major:
- 3.9: https://github.com/heroku/heroku-buildpack-python/actions/runs/18280692581
- 3.10: https://github.com/heroku/heroku-buildpack-python/actions/runs/18280679715
- 3.11: https://github.com/heroku/heroku-buildpack-python/actions/runs/18280672027
- 3.12: https://github.com/heroku/heroku-buildpack-python/actions/runs/18280667109
- 3.13: https://github.com/heroku/heroku-buildpack-python/actions/runs/18280658733

PR #1970, opened 2025-11-20
Add a check for misspelled `.python-version` files

edmorley:
We recommend that apps use a `.python-version` file to specify their Python version. However, as seen on Honeycomb after #1904, a small number of apps in the wild have spelt the file incorrectly - and sometimes with names that look very similar to the correct name, so the mistake would be easy to miss.

Some of the mistakes seen include:
- Omitting the leading `.` in the filename.
- Adding a file extension such as `.txt`.
- Using an underscore or a period instead of the hyphen.
- Having trailing whitespace in the filename.

The metrics show that some apps have both a correctly spelled file and an incorrectly spelled one, which I suspect would lead to a confusing UX if the wrong file was ever updated ("why isn't Heroku using the Python version I've specified?") - particularly in the case where the user doesn't have dotfiles visible in eg Finder and the spelling mistake was omitting the leading `.`. As such, we still error if a misspelled file is found, even if a correctly named one also exists.

GUS-W-19687420.


dzuelke:
Would it not make sense to do a case-insensitive check for the given variants?

PR #1975, opened 2025-11-21
Add more messaging to the build log recommending uv

edmorley:
The package manager [uv](https://docs.astral.sh/uv/) offers many benefits for users over the alternatives, such as it:
- Supporting lockfiles (unlike pip)
- Being much faster
- Giving more helpful error messages
- Being actively maintained by a full time team

As such we want to encourage users who may not have heard of it (or be aware that the buildpack now supports it) to try it out - and so now mention uv at the env of the build log for successful builds.

GUS-W-19124925.


spec/hatchet/pip_spec.rb
```diff
@@ -96,7 +107,7 @@
           remote: -----> Running bin/post_compile hook
           remote:        .+
           remote: -----> Saving cache
-          remote: -----> Inline app detected
```
Malax: Non blocking: Why are those lines missing now in some tests?

PR #1990, opened 2025-12-08
Convert the existing virtual environment warning to an error

edmorley:
In #1890 a warning was added if an existing Python virtual environment was found at `.venv/` or `venv/` in the root of the app dir, for the reasons described in that PR.

Now that 3 months have passed, this warning is being converted to an error.

Users with an inadvertently committed venv will need to follow the instructions in the error message to remove the venv directory from the Git index and add it to their `.gitignore` file.

GUS-W-19127259.


edmorley:
https://devcenter.heroku.com/changelog-items/3506

PR #2005, opened 2026-01-05
Remove Python 3.9 support

edmorley:
Python 3.9 reached upstream EOL on 31st October 2025:
https://devguide.python.org/versions/#supported-versions

The Python version support policy is that supported versions follows the upstream EOL lifecycle:
https://devcenter.heroku.com/articles/python-support#python-version-support-policy

And Python 3.9 support has been deprecated since 8th January 2025:
https://devcenter.heroku.com/changelog-items/3095

As such, builds of apps using Python 3.9 will now fail with an error message explaining they must be upgraded to a newer/supported Python version.

Apps using Python 3.9 that aren't able to upgrade immediately will need to pin to an older buildpack version temporarily.

GUS-W-17595553.


edmorley:
https://devcenter.heroku.com/changelog-items/3527
https://devcenter.heroku.com/articles/python-support#supported-python-versions

PR #2056, opened 2026-03-20
Add binary build support for Heroku-26

edmorley:
This adds support for building the binaries for the upcoming Heroku-26 stack based on Ubuntu 26.04.

A later PR will add support for the stack to the buildpack itself, once the Heroku build system supports Heroku-26 (since until that point we can't run Hatchet in CI anyway).

The `libcrypt-dev` library had to be added to the `Dockerfile` used for binary building, since the headers package is no longer in `heroku/heroku:26-build` as it was previously only a transitive dependency in the build image, and with the upstream Ubuntu 26.04 package changes is now no longer pulled in. (From searching it doesn't appear to be widely used, so I don't think it's worth adding back to the Heroku-26 build image.)

Doing so fixes this warning on Python 3.12 and older (the `crypt` module only exists in older Python versions):

```
*** WARNING: renaming "_crypt" since importing it failed: /tmp/src/build/lib.linux-x86_64-3.10/_crypt.cpython-310-x86_64-linux-gnu.so: undefined symbol: crypt

...

Following modules built successfully but were removed because they could not be imported:
_crypt
```

Dry-run test builds using this branch succeeded. e.g.:
https://github.com/heroku/heroku-buildpack-python/actions/runs/23343385460
https://github.com/heroku/heroku-buildpack-python/actions/runs/23343379302

GUS-W-20775546.

### DIFF
--- a/.github/workflows/build_python_runtime.yml
+++ b/.github/workflows/build_python_runtime.yml
@@ -15,6 +15,7 @@ on:
           - auto
           - heroku-22
           - heroku-24
+          - heroku-26
         default: auto
         required: false
       dry_run:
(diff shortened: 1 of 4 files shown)


runesoerensen (approved):

edmorley:
Now that Ubuntu 26.04 is GA, I've triggered Heroku-26 builds for all 71 patch version permutations. 

Note: For Python 3.10.0 to 3.10.6 this also required triggering them off a custom branch with sigstore verification disabled, since sigstore artifacts were only made available in Python 3.10.7 and newer:
https://www.python.org/downloads/metadata/sigstore/

```
for v in 3.10.{7..20} 3.11.{0..15} 3.12.{0..13} 3.13.{0..13} 3.14.{0..4}; do
    gh workflow run build_python_runtime.yml -F "python_version=${v}" -F "stack=heroku-26"
done

for v in 3.10.{0..6}; do
    gh workflow run build_python_runtime.yml -F "python_version=${v}" -F "stack=heroku-26" --ref skip-sigstore
done
```

See:
https://github.com/heroku/heroku-buildpack-python/actions/workflows/build_python_runtime.yml

PR #2076, opened 2026-04-10
Stop using global variable in `warn_or_error_if_python_version_file_missing()`

edmorley:
Longer term I'll refactor `bin/compile` to reduce the number of globals set, to prevent accidentally relying on globals in the library functions.

GUS-W-21939949.


colincasey (approved):
Nice! Makes me want to untangle a bunch of globals in the Node.js buildpack 😄

PR #2077, opened 2026-04-10
Fix `build_data::get_previous()` discarding stored `false` values

edmorley:
The jq `//` (alternative) operator treats both `null` and `false` as falsy, so a stored boolean `false` would return empty string instead of `"false"`. Fixed by replacing with an explicit `has()` check. Though in practice, no callers of `get_previously()` retrieve boolean fields, so currently nothing was affected by this.

Also  use `--arg` for the jq key parameter in both `build_data::has()` and `build_data::get_previous()`, instead of string interpolation, for consistency with `build_data::_set()`, and to avoid issues if the key name contained any special characters (though in practice they shouldn't given the names aren't user controlled).

GUS-W-21977968.
GUS-W-21977978.

### DIFF
--- a/lib/build_data.sh
+++ b/lib/build_data.sh
@@ -135,7 +135,7 @@ function build_data::_set() {
 function build_data::has() {
 	local key="${1}"
 
-	jq --exit-status "has(\"${key}\")" "${BUILD_DATA_FILE}" >/dev/null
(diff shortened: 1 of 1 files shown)


edmorley:
Once this merges, I'll port these improvements (and any others) over to the other classic buildpacks using these utility functions.

colincasey (approved):

PR #2094, opened 2026-05-11
Add support for Python 3.14.5

edmorley:
Release announcement:
https://blog.python.org/2026/05/python-3145-is-out/

Changelog:
https://docs.python.org/release/3.14.5/whatsnew/changelog.html#python-3-14-5-final

Binary builds:
https://github.com/heroku/heroku-buildpack-python/actions/runs/25693905266

GUS-W-22439823.


colincasey (approved):

edmorley:
Note: This ended up being released to buildpack registry on 2026-06-04 and not 2026-05-11, since the original v345 release prepared in #2095 was inadvertently not published after merge, so the change wasn't released until #2100.

The automation added since then in #2099 removes this manual step, which will prevent this from happening in the future.

PR #2099, opened 2026-06-04
Add OIDC-based buildpack release workflow

colincasey:
Calls the shared `_classic-buildpack-publish.yml` workflow from `heroku/languages-github-actions` to publish the `heroku/python` buildpack to the registry via OIDC.

Auto-triggers when the Linguist-authored "Prepare release" PR is merged to `main` (guarded on event, branch, repo, and PR author). Also supports manual `workflow_dispatch`.

Mirrors the pattern established in heroku/heroku-buildpack-nodejs#1653.

W-22295409 is the reference implementation; this is the python caller workflow for the OIDC-Based Classic Buildpack Publishing epic.

W-22295419

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,40 @@
+name: Release
+
+on:
+  # Auto-trigger when the "Prepare release" PR (created by the
+  # _classic-buildpack-prepare-release.yml workflow) is merged.
+  pull_request:
+    branches:
+      - main
+    types:
+      - closed
+  workflow_dispatch:
+    inputs:
+      qa:
+        description: "QA mode (staging registry, -qa tag suffix, draft releases)"
+        required: true
+        default: false
+        type: boolean
+
+permissions:
+  id-token: write
+  contents: write
+
+jobs:
+  release:
+    name: Release
+    # On `pull_request`, only run for the merged auto-generated
+    # "Prepare release" PR (branch name set by
+    # _classic-buildpack-prepare-release.yml). Manual dispatches always run.
+    if: >-
+      github.event_name == 'workflow_dispatch' ||
+      (github.event.pull_request.merged == true &&
+       github.event.pull_request.head.repo.full_name == github.repository &&
+       github.event.pull_request.head.ref == 'prepare-release' &&
+       github.event.pull_request.user.login == 'heroku-linguist[bot]')
+    uses: heroku/languages-github-actions/.github/workflows/_classic-buildpack-publish.yml@latest
+    with:
+      buildpack_id: "heroku/python"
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,40 @@
...
+      - main
+    types:
+      - closed
+  workflow_dispatch:
+    inputs:
+      qa:
+        description: "QA mode (staging registry, -qa tag suffix, draft releases)"
+        required: true
+        default: false
+        type: boolean
```
edmorley: I don't think we'll ever have a need to do a QA run from most buildpacks - I think I'd prefer to omit this from all of these workflows added to the classic buildpacks, and just leave it as an optional thing in the reusable workflow that can be enabled from a test buildpack iff needed?

.github/workflows/release.yml
```diff
@@ -0,0 +1,40 @@
+name: Release
```
edmorley: Could we call this "Release Buildpack" for consistency with the CNB release workflows?

eg:
https://github.com/heroku/buildpacks-python/blob/main/.github/workflows/release.yml#L1

.github/workflows/release.yml
```diff
@@ -0,0 +1,40 @@
...
+      (github.event.pull_request.merged == true &&
+       github.event.pull_request.head.repo.full_name == github.repository &&
+       github.event.pull_request.head.ref == 'prepare-release' &&
+       github.event.pull_request.user.login == 'heroku-linguist[bot]')
+    uses: heroku/languages-github-actions/.github/workflows/_classic-buildpack-publish.yml@latest
+    with:
+      buildpack_id: "heroku/python"
+      # `inputs.qa` is unset on `pull_request`, so coerce to false — auto-triggered
+      # publishes always go to production. Manual dispatches pass the input through.
+      qa: ${{ inputs.qa || false }}
```
edmorley: This can also be removed it we omit the QA `inputs` above.

edmorley (changes requested):

.github/workflows/release.yml
colincasey: Renamed to "Release Buildpack".

.github/workflows/release.yml
colincasey: Agreed — dropped the QA `workflow_dispatch` input. QA staging publishes stay an optional feature of the reusable `_classic-buildpack-publish.yml` (the `qa` input is `required: false`, defaults to `false`), so it can be enabled from a test buildpack iff needed rather than exposed on every classic buildpack caller.

.github/workflows/release.yml
colincasey: Removed along with the `qa` input.

edmorley (approved):
Thank you!

